PR #40, opened 2015-01-08
Allow for a globally-configurable 'pending' timeout

sschneid:


sschneid:
I've verified this as working in development.

PR #41, opened 2015-01-09
Rename 'template'/'count' to 'key'/'value'

sschneid:
This should make the code cleaner; things other than a template name will be used in the POST interface.


sschneid:
This has been tested in dev.

PR #42, opened 2015-01-13
As-is, TTL-cleanup is an hour off

sschneid:
This commit changes calculation from "greater than" to "greater than or equal to" TTL.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -146,7 +146,7 @@ module Vmpooler
             if (
               (host.runtime) and
               (host.runtime.bootTime)
-              ((((Time.now - host.runtime.bootTime)/60).to_s[/^\d+\.\d{1}/].to_f) > ttl)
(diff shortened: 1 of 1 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -146,7 +146,7 @@ def check_running_vm vm, pool, ttl
             if (
               (host.runtime) and
               (host.runtime.bootTime)
-              ((((Time.now - host.runtime.bootTime)/60).to_s[/^\d+\.\d{1}/].to_f) > ttl)
+              ((((Time.now - host.runtime.bootTime)/60).to_s[/^\d+\.\d{1}/].to_f) >= ttl)
```
rji: I understand the diff just fine, but I'm curious why you need to convert to a string and then back to a float.

branan: looks like `floor()` implemented via regex?

PR #43, opened 2015-01-27
Record clone times in Redis, add a '/status' API endpoint

sschneid:


lib/vmpooler/api.rb
```diff
@@ -168,6 +168,45 @@ def hostname_shorten hostname
...
+          result = {}
+          clone_times = []
+
+          result['capacity_current'] = 0
+          result['capacity_total'] = 0
+
+          result['status'] = 1
+
+          $config[:pools].each do |pool|
+            result['capacity_current'] += $redis.scard( 'vmpooler__ready__' + pool['name'] ).to_i
```
colinPL: If this redis key is missing, what happens? Does `capacity_current` retain the original `0` value or does this just throw an error and return a `5xx` status?

lib/vmpooler/api.rb
```diff
@@ -168,6 +168,45 @@ def hostname_shorten hostname
...
+            result['capacity_total'] += pool['size'].to_i
+
+            if ( $redis.scard( 'vmpooler__ready__' + pool['name'] ).to_i == 0 )
```
colinPL: Can we use `result['capacity_current']` here instead of hitting redis again?

sschneid: Good call.

lib/vmpooler/api.rb
sschneid: If the key happens to be missing, Redis will return a '0' integer and result['capacity_current'] will remain the same value.

lib/vmpooler/api.rb
```diff
@@ -168,6 +168,45 @@ def hostname_shorten hostname
           JSON.pretty_generate(result)
         end
 
+        get '/status/?' do
+          content_type :json
+
+          result = {}
+          clone_times = []
```
colinPL: Is this endpoint being used for health and monitoring purposes? Would it be beneficial to have clone times stored with the hostname that was cloned (a hash instead of an array)?

lib/vmpooler/api.rb
sschneid: Actually, in hindsight, 'current_capacity' is OVERALL throughout all of the pools.  This check is just to see if the existing pool is empty.

lib/vmpooler/api.rb
sschneid: I'd say "not at this point", no... perhaps if our monitoring evolves in the future to need more granular clone-time judgement, but as I envision it currently... no.

PR #44, opened 2015-02-03
Store and retrieve daily clone statistics from Redis db

sschneid:


lib/vmpooler/api.rb
```diff
@@ -189,23 +188,17 @@ def hostname_shorten hostname
...
             result['status'] = 0
           end
 
-          if ( clone_times.any? )
-            result['clone_average'] = clone_times.map( &:to_f ).reduce( :+ ) / clone_times.size
-          end
-
           result['capacity_perecent'] = ( result['capacity_current'].to_f / result['capacity_total'].to_f ) * 100.0
 
+          result['clone_total'] = $redis.hlen('vmpooler__clone__'+Date.today.to_s)
```
colinPL: This is total for today, not for 24 hours. I do not see a piece that calculates `now - 24 hours`.

sschneid: Nobody said in this PR it was for 24-hours :)

lib/vmpooler/api.rb
```diff
@@ -189,23 +188,17 @@ def hostname_shorten hostname
...
+          result['clone_average'] = $redis.hvals('vmpooler__clone__'+Date.today.to_s).map( &:to_f ).reduce( :+ ) / result['clone_total']
```
colinPL: Should this have any guards incase `result['clone_total']` is `0`?

sschneid: Patched

PR #45, opened 2015-02-04
Display application 'uptime' in API status endpoint

sschneid:


lib/vmpooler/api.rb
```diff
@@ -201,6 +203,8 @@ def hostname_shorten hostname
             result['clone_average'] = $redis.hvals('vmpooler__clone__'+Date.today.to_s).map( &:to_f ).reduce( :+ ) / result['clone_total']
           end
 
+          result['uptime'] = Time.now - $config[:uptime]
```
rji: Ideally, what units should this be? Seconds?

sschneid: Yes, 'clone_average' and now 'status' are both in seconds.  I'll keep in mind to maintain consistency throughout this endpoint.

lib/vmpooler/api.rb
```diff
@@ -5,6 +5,8 @@ def initialize
       config_file = File.expand_path('vmpooler.yaml')
       $config = YAML.load_file(config_file)
 
+      $config[:uptime] = Time.now
```
rji: So this comment is me not knowing what Ruby's `Time.now` is really going to do.

When I look at Time.now, irb defaults to displaying it as a string, which corrects for DST. However, when I do math with it, it seems to know what's up. In Python, time is extremely verbose. Do you know how this will behave around DST, or would it be safer to do convert it to an integer, where we can guarantee it'll be something like:

```
$ date +%s
1423083750
$ irb
>> Time.now.to_i
=> 1423083756
>> 
$ 
```

sschneid: It is a Ruby 'time' object; print it displays the string, operations done on it (as I'm doing) are performed on the integer value and returned as seconds.

http://www.ruby-doc.org/core-2.2.0/Time.html#method-i-2D

rji: :+1: thanks for the clarification

PR #47, opened 2015-02-06
Rubocop --auto-correct syntax fixups

sschneid:


lib/vmpooler/api.rb
```diff
@@ -8,25 +8,24 @@ def initialize
...
       $config[:redis]['server'] ||= 'localhost'
 
-      if ($config[:graphite]['server'])
+      if $config[:graphite]['server']
         $config[:graphite]['prefix'] ||= 'vmpooler'
       end
 
       # Connect to Redis
-      $redis = Redis.new(:host => $config[:redis]['server'])
+      $redis = Redis.new(host: $config[:redis]['server'])
```
colinPL: Is `host` still a symbol here?

rji:
Seems legit. :+1:

lib/vmpooler/api.rb
sschneid: I guess?  I deployed the reformatted code to dev and everything seems to function identically AFAICT.

lib/vmpooler/api.rb
```diff
@@ -292,12 +291,12 @@ def hostname_shorten hostname
...
 
           request.keys.each do |template|
-            if ( $redis.scard('vmpooler__ready__'+template) < request[template] )
+            if $redis.scard('vmpooler__ready__' + template) < request[template]
               available = 0
             end
           end
 
-          if ( available == 1 )
+          if (available == 1)
```
colinPL: Why did it keep the parens here?

colinPL:
:+1:

PR #48, opened 2015-02-20
Add /summary endpoint

colinPL:
Add an endpoint that returns clone totals and average clone time for
the given time spans. The time spans are determined by the query
parameters "from" and "to", with the resulting range being inclusive of
both. If you want a single day, it should be both the "from" and the
"to" value.

### DIFF
--- a/lib/vmpooler/api.rb
+++ b/lib/vmpooler/api.rb
@@ -31,6 +31,10 @@ module Vmpooler
 
             hostname
           end
+
+          def validate_date_str(date_str)
+            /^\d{4}-\d{2}-\d{2}$/ === date_str
+          end
         end
 
         get '/' do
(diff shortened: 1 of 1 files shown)


lib/vmpooler/api.rb
```diff
@@ -31,6 +31,10 @@ def hostname_shorten(hostname)
 
             hostname
           end
+
+          def validate_date_str(date_str)
+            /^\d{4}-\d{2}-\d{2}$/ === date_str
```
rji: Is there a reason we're not just using `DateTime.iso8601()`?

```
>> DateTime.iso8601('02-15-2015')
ArgumentError: invalid date
	from (irb):8:in `iso8601'
	from (irb):8
	from /Users/roger/.rvm/rubies/ruby-2.1.2/bin/irb:11:in `<main>'
>> DateTime.iso8601('2015-02-15')
=> #<DateTime: 2015-02-15T00:00:00+00:00 ((2457069j,0s,0n),+0s,2299161j)>
```

colinPL: I was following the format that is displayed when you execute `Date.today.to_s`

PR #54, opened 2015-02-24
(Re)structured /status API endpoint

sschneid:


lib/vmpooler/api.rb
```diff
@@ -157,53 +157,72 @@ def mean(list)
...
             end
 
-            result[:pending] += $redis.scard('vmpooler__pending__' + pool['name'])
-            result[:ready] += $redis.scard('vmpooler__ready__' + pool['name'])
-            result[:running] += $redis.scard('vmpooler__running__' + pool['name'])
-            result[:completed] += $redis.scard('vmpooler__completed__' + pool['name'])
+            result[:queue][:pending] += $redis.scard('vmpooler__pending__' + pool['name'])
+            result[:queue][:ready] += $redis.scard('vmpooler__ready__' + pool['name'])
+            result[:queue][:running] += $redis.scard('vmpooler__running__' + pool['name'])
+            result[:queue][:completed] += $redis.scard('vmpooler__completed__' + pool['name'])
```
colinPL: This and 3 lines above should end in `.to_i`.

lib/vmpooler/api.rb
```diff
@@ -157,53 +157,72 @@ def mean(list)
...
 
-          if result[:empty]
-            result[:status] = 0
+          if result[:status][:empty]
+            result[:status][:ok] = false
+            result[:status][:message] = "Found #{result[:status][:empty].length} empty pools."
           end
 
-          result[:capacity_percent] = ((result[:capacity_current].to_f / result[:capacity_total].to_f) * 100.0).round(1)
+          result[:capacity][:percent] = ((result[:capacity][:current].to_f / result[:capacity][:total].to_f) * 100.0).round(1)
```
colinPL: This should probably verify that `result[:capacity][:total].to_f` is  not `0`.

lib/vmpooler/api.rb
```diff
@@ -157,53 +157,72 @@ def mean(list)
...
+          result[:queue][:cloning] = $redis.get('vmpooler__tasks__clone').to_i
+          result[:queue][:booting] = result[:queue][:pending].to_i - result[:queue][:cloning].to_i
+          result[:queue][:booting] = 0 if result[:queue][:booting] < 0
+          result[:queue][:total] = result[:queue][:pending].to_i + result[:queue][:ready].to_i + result[:queue][:running].to_i + result[:queue][:completed].to_i
 
-          result[:cloning] = $redis.get('vmpooler__tasks__clone')
-          result[:booting] = result[:pending].to_i - result[:cloning].to_i
-          result[:booting] = 0 if result[:booting] < 0
-          result[:total] = result[:pending].to_i + result[:ready].to_i + result[:running].to_i + result[:completed].to_i
+          result[:clone][:count][:total] = $redis.hlen('vmpooler__clone__' + Date.today.to_s)
```
colinPL: This could use `.to_i`.

lib/vmpooler/api.rb
```diff
@@ -157,53 +157,72 @@ def mean(list)
...
+          if result[:clone][:count][:total] > 0
+            clone_times = $redis.hvals('vmpooler__clone__' + Date.today.to_s).map(&:to_f)
 
-          result[:clone_total] = $redis.hlen('vmpooler__clone__' + Date.today.to_s)
-          if result[:clone_total] > 0
-            result[:clone_average] = ($redis.hvals('vmpooler__clone__' + Date.today.to_s).map(&:to_f).reduce(:+) / result[:clone_total]).round(1)
+            result[:clone][:duration][:average] = (clone_times.reduce(:+) / result[:clone][:count][:total]).round(1)
```
colinPL: `reduce(:+)` on an empty list will return a `nil`:

``` ruby
irb(main):001:0> [].map(&:to_f)
=> []
irb(main):002:0> [].map(&:to_f).reduce(:+)
=> nil
```

lib/vmpooler/api.rb
colinPL: `percent` is not set in initial `result` creation.

lib/vmpooler/api.rb
```diff
@@ -157,53 +157,72 @@ def mean(list)
...
+            result[:status][:message] = "Found #{result[:status][:empty].length} empty pools."
           end
 
-          result[:capacity_percent] = ((result[:capacity_current].to_f / result[:capacity_total].to_f) * 100.0).round(1)
+          result[:capacity][:percent] = ((result[:capacity][:current].to_f / result[:capacity][:total].to_f) * 100.0).round(1)
+
+          result[:queue][:cloning] = $redis.get('vmpooler__tasks__clone').to_i
+          result[:queue][:booting] = result[:queue][:pending].to_i - result[:queue][:cloning].to_i
+          result[:queue][:booting] = 0 if result[:queue][:booting] < 0
+          result[:queue][:total] = result[:queue][:pending].to_i + result[:queue][:ready].to_i + result[:queue][:running].to_i + result[:queue][:completed].to_i
```
colinPL: This is not set in initial `result` creation.

lib/vmpooler/api.rb
sschneid: 0.0 is a valid percent.

```
  "capacity": {
    "current": 0,
    "total": 10,
    "percent": 0.0
  }```

sschneid:
I think I've addressed all of the above.

PR #56, opened 2015-02-25
Use today as a default 'from' value

sschneid:


lib/vmpooler/api.rb
```diff
@@ -258,8 +258,10 @@ def mean(list)
             halt 400, 'Invalid "to" date format, must match YYYY-MM-DD'
           end
 
-          # We do need a from date, however.
-          if from_param.nil? or !validate_date_str(from_param.to_s)
+          # It is also ok if from_param is nil, we will assume this also means 'today'
```
rji: I would assume that the `from` param always has to be some date in the past, and that the `to` param could be some date in the past _or_ 'today', if it isn't specified. We already assume that an empty `to` param is 'today' on lines 255-256.

What is the benefit of assuming that the `from` date is 'today' if the param isn't specified?

sschneid: It allows the param to be unset, and the default `/summary` view (with no params supplied) is a summary of the current day.

colinPL: Currently, a missing `from` value will result in a 40x error code. This change will allow a request to `/summary` without any parameters required and return today's information.

colinPL:
:+1: .. awaiting response from @rji

rji:
I think this might be cleaner if it was implemented as multiple Sinatra routes, say `/summary` and `/summary/?`, and each of those called a method that implemented this code, instead of having the code live in the route itself.

But since we don't do this anywhere else yet, I'm not going to block the PR on my comments alone. :+1:

PR #57, opened 2015-02-25
Set Sinatra not_found to JSON ok: false

sschneid:


lib/vmpooler/api.rb
```diff
@@ -42,6 +42,16 @@ def mean(list)
           end
         end
 
+        not_found do
```
colinPL: Does this still send a `404` HTTP code?

sschneid: Nope.  I was going to audit and set valid status codes in a different PR.

PR #58, opened 2015-02-25
(maint) Fix Divide by 0 Bug

colinPL:
When a redis key does not exist for a single-day summary, the
total_clones_per_day array has a length of 1 with a single value of 0.
This gets past an initial check of array length, but is then reduced to
a value of 0, which is then used as the denominator of an average
calculation. The result of this calculation is NaN and later causes
JSON.pretty_generate to raise an exception.

This also includes a check on the min_max_clone_times array, because
[].minmax returns [nil, nil].


sschneid:
:+1:

PR #59, opened 2015-02-26
Set HTTP status codes, a bit of code standardization

sschneid:
- set proper HTTP status codes
- move 'content_type :json' to the beginning of endpoint defs
- deprecate 'Hash.new' for '{}'

### DIFF
--- a/lib/vmpooler/api.rb
+++ b/lib/vmpooler/api.rb
@@ -207,7 +207,6 @@ module Vmpooler
           end
 
           if result[:status][:empty]
-            status 503
             result[:status][:ok] = false
             result[:status][:message] = "Found #{result[:status][:empty].length} empty pools."
           end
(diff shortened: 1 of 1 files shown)


colinPL:
I do not see a status code being set for `not_found`. Does that already return a 404 status code?

sschneid:
> Does that already return a 404 status code?

Yup.

lib/vmpooler/api.rb
```diff
@@ -205,6 +207,7 @@ def mean(list)
           end
 
           if result[:status][:empty]
+            status 503
```
colinPL: Returning a `503` says that the request to this URL had an issue and could not be processed. I think you are saying that the pooler itself is in a bad state - not the HTTP request itself - and I think the `result[:status][:ok] = false` relays that information.

I would like this to return a `200 OK` regardless of the pooler's pool status.

PR #62, opened 2015-03-04
Metric calc via helpers

sschneid:
This PR moves common metrics calculations to helpers.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -31,7 +31,8 @@ module Vmpooler
           duration: {
             average: 0,
             min: 0,
-            max: 0
+            max: 0,
+            total: 0
           },
           count: {
             total: 0
@@ -43,7 +44,8 @@ module Vmpooler
         if clone[:count][:total] > 0
           clone_times = get_clone_times(date)
 
-          clone[:duration][:average] = (clone_times.reduce(:+).to_f / clone[:count][:total]).round(1)
+          clone[:duration][:total] = clone_times.reduce(:+).to_f
+          clone[:duration][:average] = (clone[:duration][:total] / clone[:count][:total]).round(1)
           clone[:duration][:min], clone[:duration][:max] = clone_times.minmax
         end
 
@@ -146,7 +148,7 @@ module Vmpooler
                   total: 0
               }
           },
-          daily: []   # used for daily info
+          daily: []
       }
 
       from_param = params[:from] || Date.today.to_s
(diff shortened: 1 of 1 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -5,48 +5,59 @@ class V1 < Sinatra::Base
...
+            total: 0
           },
           count: {
             total: 0
           }
-        },
-        queue: {
+        }
+
+        clone[:count][:total] = $redis.hlen('vmpooler__clone__' + date).to_i
```
colinPL: Is `date` a date string or an actual date? If the latter, this needs `.to_s`. If the former, then I suggest renaming the variable to `date_str`. (methods below use "date" and treat it as a date object)

sschneid: Anything that uses one of the helper functions should be passing the `date` param as a string.

lib/vmpooler/api/v1.rb
```diff
@@ -5,48 +5,59 @@ class V1 < Sinatra::Base
...
+          clone[:duration][:total] = clone_times.reduce(:+).to_f
+          clone[:duration][:average] = (clone[:duration][:total] / clone[:count][:total]).round(1)
+          clone[:duration][:min], clone[:duration][:max] = clone_times.minmax
+        end
+
+        clone
+      end
+
+      def get_clone_times(date)
+        $redis.hvals('vmpooler__clone__' + date.to_s).map(&:to_f)
```
colinPL: `date` is already a string.

lib/vmpooler/api/v1.rb
sschneid: (Updated anyway)

lib/vmpooler/api/v1.rb
```diff
@@ -117,109 +148,56 @@ def mean(list)
...
-          # this prevents divide by 0 problems.
-          if me[:clone][:count][:total] > 0
-            me[:clone][:duration][:average] = mean(clone_times)
-          else
-            me[:clone][:duration][:average] = 0
-          end
-        end
+      result[:daily].each do |daily|
+        daily_clone_counts.push(daily[:clone][:count][:total])
+        daily_clone_durations.push(daily[:clone][:duration][:min])
```
colinPL: If this `daily` did not clone anything (`[:count][:total] == 0`), then `min` and `max` will be the default of `0`. Appending those values to `daily_clone_durations` will cause future processing to select `0` as the overall minimum duration for this summary. These values should only be appended if the `daily` cloned something.

colinPL:
Once the min/max comment is addressed, I am :+1:

lib/vmpooler/api/v1.rb
sschneid: Addressed in https://github.com/sschneid/vmpooler/commit/e347e2b6b4337d78dfc9dab908e09584a0bb3c5a.

PR #65, opened 2015-03-06
/summary per-pool metrics

sschneid:
This commit also consolidates the 'get_boot_metrics' and
'get_clone_metrics' methods into a single parameterized
'get_task_metrics' method.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -5,36 +5,6 @@ module Vmpooler
     api_prefix  = "/api/v#{api_version}"
 
     helpers do
-      def get_boot_metrics(date_str)
-        boot = {
-          duration: {
-            average: 0,
-            min: 0,
-            max: 0,
-            total: 0
-          },
-          count: {
-            total: 0
-          }
-        }
-
-        boot[:count][:total] = $redis.hlen('vmpooler__boot__' + date_str).to_i
-
-        if boot[:count][:total] > 0
-          boot_times = get_boot_times(date_str)
-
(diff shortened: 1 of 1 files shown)


sschneid:
This is deployed on vmpooler-dev.  Verification of my maths is probably needed.

lib/vmpooler/api/v1.rb
```diff
@@ -240,13 +270,25 @@ def validate_date_str(date_str)
           result[task][:duration][:total] += daily[task][:duration][:total]
 
           if result[task][:count][:total] > 0
```
colinPL: This block should be outside the loop so that the total duration average is only calculated once, not altered each day.

lib/vmpooler/api/v1.rb
```diff
@@ -240,13 +270,25 @@ def validate_date_str(date_str)
...
         result[task][:count][:min], result[task][:count][:max] = daily_counts.minmax
         result[task][:count][:average] = mean(daily_counts)
         result[task][:duration][:min], result[task][:duration][:max] = daily_durations.minmax
+
+        daily_counts_bypool.each_key do |pool|
+          result[task][:count][:pool][pool][:min], result[task][:count][:pool][pool][:max] = daily_counts_bypool[pool].minmax
+          result[task][:count][:pool][pool][:average] = mean(daily_counts_bypool[pool])
+
+          result[task][:duration][:pool][pool] ||= {}
+          result[task][:duration][:pool][pool][:min], result[task][:duration][:pool][pool][:max] = daily_durations_bypool[pool].minmax
```
colinPL: If `daily_duration` is empty, then `min` and `max` will be `nil`.

PR #66, opened 2015-03-09
Use $redis.hgetall rather than hget in a loop

sschneid:
This cuts down on render-time for the /api/v1/summary endpoint drastically.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -76,7 +76,7 @@ module Vmpooler
             task[:count][:pool] = {}
             task[:duration][:pool] = {}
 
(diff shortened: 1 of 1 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -91,6 +91,7 @@ def get_task_metrics(task_str, date_str, opts = {})
 
               task_times_bypool[pool] ||= []
               task_times_bypool[pool].push($redis.hget('vmpooler__' + task_str + '__' + date_str, key).to_f)
+              task_times_bypool[pool].push(value.to_f)
```
rji: With the move to `hgetall()`, do you still need line 93?

sschneid: No.  Oops.

rji:
:+1:

PR #68, opened 2015-03-11
Allow authorized_keys to be overridden on VM checkout

sschneid:


sschneid:
This was a POC; closing until there is a demonstrated / ticketed need for this.

PR #70, opened 2015-03-19
(QENG-1905) Adding VM-tagging support via PUT /vm/:hostname endpoint

sschneid:


sschneid:
Deployed on vmpooler-dev (internal).

rji:
If I understand correctly, this will allow arbitrary tags to be set, but they'll only persist in Redis for the lifetime of the VM. Could we also ship these to Graphite for historical usage reporting?

README.md
```diff
@@ -168,6 +168,15 @@ $ curl --url vmpooler.company.com/vm/pxpmtoonx7fiqg6
...
 Modify a checked-out VM.
 
+The following are valid PUT parameters and their required data structures:
+
+parameter | description | required structure
+--------- | ----------- | ------------------
+*lifetime* | VM TTL (in hours) | integer
+*tags* | free-form VM tagging | hash
+
+Any modifications can be verified using the GET /vm/&lt;hostname&gt endpoint.
```
rji: Missing a semicolon after `&gt`

sschneid:
> If I understand correctly, this will allow arbitrary tags to be set, but they'll only persist in Redis for the lifetime of the VM. Could we also ship these to Graphite for historical usage reporting?

That's correct.  I'm not sure about Graphite though, since it's not really time-series data...

rji:
Hrm, yeah good point. I know we log when a VM was checked out, what if we appended the tags (if any) to that line in the log file, then ship those logs to our ELK cluster? It would still allow us to use arbitrary tags, and not add too much more maintenance/overhead to vmpooler.

sschneid:
Redis data persistence addressed in #71.

lib/vmpooler/api/v1.rb
```diff
@@ -506,6 +515,15 @@ def validate_date_str(date_str)
                 status 200
                 result['ok'] = true
               end
+            when 'tags'
+              if arg.is_a?(Hash)
```
colinPL: What if `arg` is not a hash? Should this set a status of `400` (bad request) and bail? It is possible - because of the `.each` - that a request with both `lifetime` and `tags` can return a `200 OK`, but actually not record the tags or lifetime, depending on the order of arguments and how they are parsed in the `each` block.

sschneid: Good catch... I'll close this and re-open with a fixup.

lib/vmpooler/api/v1.rb
```diff
@@ -496,17 +505,32 @@ def validate_date_str(date_str)
...
+                failure = true
               end
           end
+
+          if failure
+            status 400
+          else
+            status 200
+            result['ok'] = true
         end
```
colinPL: Missing an `end`. Also, this `if` is within the `.each` block. This needs to be after that block, otherwise a successful param will overwrite `failure` to `false` at the top.

sschneid: Grrrrrrr...

sschneid:
Reopened as #72.

PR #71, opened 2015-03-20
(QENG-2057) Historic Redis VM metadata

sschneid:
- Store VM 'checkout' and 'destroy' timestamps in Redis VM metadata object
- Do not destroy said object upon VM destruction
- Implement a Vmpooler::Janitor worker thread which compares the 'destroy' timestamp with a configurable redis['data_ttl'] option; destroy metadata object if TTL is exceeded

### DIFF
--- a/lib/vmpooler/janitor.rb
+++ b/lib/vmpooler/janitor.rb
@@ -1,0 +1,39 @@
+module Vmpooler
+  class Janitor
+    def initialize
+      # Load the configuration file
+      config_file = File.expand_path('vmpooler.yaml')
+      $config = YAML.load_file(config_file)
+
+      # Set some defaults
+      $config[:redis]             ||= {}
+      $config[:redis]['server']   ||= 'localhost'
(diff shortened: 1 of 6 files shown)


lib/vmpooler/janitor.rb
```diff
@@ -0,0 +1,39 @@
+module Vmpooler
+  class Janitor
+    def initialize
+      # Load the configuration file
+      config_file = File.expand_path('vmpooler.yaml')
+      $config = YAML.load_file(config_file)
```
rji: This code is repeated a few times throughout the app. Perhaps it would be better suited in its own class or helper method?

sschneid: Agreed... I think there are likely several chunks of code that would be better suited as helper methods.  I think modifying all of the libraries would be out of scope for this PR, though?  I'd rather do it in it's own individual PR, at least; what do you think?

rji: You're right; a new ticket/PR for that makes sense to me.

PR #77, opened 2015-03-27
(QENG-1906) Refactor initialize to allow config passing

colinPL:
Prior to this commit, several pieces of vmpooler performed configuration
and initialization steps within 'initialize'. This made it difficult to
pass in mock objects for testing purposes.

This commit performs a single configuration and passes the results to
the various pieces of vmpooler.

### DIFF
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -18,4 +18,47 @@ module Vmpooler
       require File.expand_path(File.join(File.dirname(__FILE__), 'vmpooler', lib))
     end
   end
+
+  def self.config(filepath='vmpooler.yaml')
+    # Load the configuration file
+    config_file = File.expand_path(filepath)
+    parsed_config = YAML.load_file(config_file)
+
+    # Set some defaults
+    parsed_config[:redis]             ||= {}
+    parsed_config[:redis]['server']   ||= 'localhost'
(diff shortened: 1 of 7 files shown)


rji:
I think these commits should be squashed; the last two seem to be fixups for the first.

rji:
I'd like to put this through its paces on the dev environment (if that hasn't been done already, of course). But otherwise the diffs look sane enough to me.

sschneid:
This branch has been deployed to and is currently running on vmpooler-dev.

lib/vmpooler.rb
```diff
@@ -18,4 +18,47 @@ module Vmpooler
       require File.expand_path(File.join(File.dirname(__FILE__), 'vmpooler', lib))
     end
   end
+
+  def self.config
+    # Load the configuration file
+    config_file = File.expand_path('vmpooler.yaml')
+    $config = YAML.load_file(config_file)
```
rji: Is there a reason we still need to use global variables with all of this in `self.config` now?

colinPL: I did not look at changing the variable's type when relocating the logic. Should this be changed or can it wait?

rji: I think we should change it to local scope and make sure nothing else breaks

PR #78, opened 2015-03-30
(QENG-1906) Add spec tests for Janitor

colinPL:
This adds tests for Janitor, verifying it deletes stale VMs.

**NOTE: THIS IS DEPENDENT ON PR #77**

### DIFF
--- a/spec/vmpooler/janitor_spec.rb
+++ b/spec/vmpooler/janitor_spec.rb
@@ -1,0 +1,71 @@
+require 'spec_helper'
+
+describe 'Janitor' do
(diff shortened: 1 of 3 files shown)


spec/vmpooler/janitor_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'rspec'
+require 'vmpooler'
```
rji: You should `require 'spec_helper'` here instead

sschneid:
:+1:

Issue #81, opened 2015-03-31
NilClass error when running API without Graphite configured


fatmcgav:
If you attempt to use the API without Graphite configured, you get a 'undefined method [] for NilClass' error, due to [api.rb#L14](https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/api.rb#L14)

PR incoming to fix :)

mattkirby:
I believe this is resolved with the addition of a dummy stats provider.

PR #82, opened 2015-03-31
Fix for #81 to check if graphite server is defined in config

fatmcgav:
Fixes issue #81.

### DIFF
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -33,7 +33,7 @@ module Vmpooler
     parsed_config[:config]['vm_checktime'] ||= 15
     parsed_config[:config]['vm_lifetime']  ||= 24
 
-    if parsed_config[:graphite]['server']
+    if parsed_config.has_key?(:graphite)
       parsed_config[:graphite]['prefix'] ||= 'vmpooler'
     end
 


colinPL:
This needs a rebase.

fatmcgav:
@colinPL Will take a look now :)

fatmcgav:
@colinPL Actually, looks like it's already been fixed here: https://github.com/puppetlabs/vmpooler/blob/fd90756b834dc6f31fa066abffd85312aed3f882/vmpooler#L12

Will give that a go tomorrow... :)

fatmcgav:
I was still seeing nil class errors after pulling the latest changes in, so have tweaked the commit.

lib/vmpooler.rb
```diff
@@ -33,7 +33,7 @@ def self.config(filepath='vmpooler.yaml')
     parsed_config[:config]['vm_checktime'] ||= 15
     parsed_config[:config]['vm_lifetime']  ||= 24
 
-    if parsed_config[:graphite]['server']
+    if defined? parsed_config[:graphite]['server']
```
colinPL: Do you know when `defined?` returns `nil` or `false`? Below it always returns a string, except when applied to an actually undefined variable (`k`).

``` ruby
irb(main):002:0> t = {graphite: {'server' => 'blah'}}
=> {:graphite=>{"server"=>"blah"}}
irb(main):003:0> defined? t
=> "local-variable"
irb(main):004:0> defined? k
=> nil
irb(main):005:0> defined? t[:graphite]['server']
=> "method"
irb(main):006:0> defined? t[:graphite]['asdf']
=> "method"
irb(main):007:0> defined? t[:graphite]
=> "method"
irb(main):008:0> defined? t[:graph]
=> "method"
```

colinPL: I think you want `parsed_config.has_key?(:graphite)`. That will check that the `:graphite` key even exists in the config, which will be `false` in your case.

fatmcgav:
Reworked it... 

Checking for top-level `:graphite` key being set, but should it also check for or fail if `[:graphite][:server]` isn't set?

colinPL:
@sschneid - thoughts?

sschneid:
I think that checking for both `[:graphite]` and `[:graphite][:server]` makes sense since they'll both need to be set if Graphite support is going to work...

sschneid:
I think that this fix is working, but I'm encountering a similar/related new (doh) error if I disable graphite while testing:

````
NoMethodError: undefined method `[]' for nil:NilClass
  (root) at vmpooler:11
````

I'm going to label this PR as 'blocked' until we can get a fix in so that this can be properly tested.

colinPL:
Closing due to inactivity.

PR #85, opened 2015-04-01
Remove defined? when checking configuration for graphite server.

colinPL:
This is a quick patch to solve the issue of events no longer reporting to graphite. This was due to an improper use of `defined?`.


sschneid:
:+1:

PR #86, opened 2015-04-01
(QENG-1906) Break apart check_pending_vm and add spec tests

colinPL:
This commit breaks apart the method check_pending_vm in to three,
smaller methods that are easier to test. There are no functionality
changes or bug fixes.

This adds spec tests for the overall functionality that was
check_pending_vm.


sschneid:
:+1:  I've manually acceptance-tested this.  Looks good.

PR #88, opened 2015-04-02
(QENG-1906) Add specs for Dashboard and root API class

colinPL:
Add spec tests for Sinatra powered Dashboard piece. This includes tests
for the homepage (/) and a nonexistent route.


colinPL:
This will require the `rack-test` gem to be installed:

```
$ gem install rack-test
```

sschneid:
> This requires the `rack-test` gem to be installed

:disappointed:  :cry: :sob: :panda_face:

PR #91, opened 2015-04-09
Ensure 'lifetime' val returned by GET /vm/:hostname is an int

sschneid:
Prior to this PR, if a VM's `lifetime` was changed via `POST /vm/:hostname`, the value pulled from Redis is evaluated as a string.  This forced it to return an integer in the endpoint output.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -452,7 +452,7 @@ module Vmpooler
         result[params[:hostname]] = {}
 
         result[params[:hostname]]['template'] = rdata['template']
-        result[params[:hostname]]['lifetime'] = rdata['lifetime'] || Vmpooler::API.settings.config[:config]['vm_lifetime']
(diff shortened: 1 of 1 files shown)


sschneid:
Tested on vmpooler-dev.

````
$ curl --url vmpooler-dev/vm/gj3hpatv4qf9kfn
{
  "ok": true,
  "gj3hpatv4qf9kfn": {
    "template": "centos-6-x86_64",
    "lifetime": 1,
    "running": 0,
  }
}
````
````
$ curl -X PUT -d '{"lifetime":"2"}' --url vmpooler-dev/vm/gj3hpatv4qf9kfn
{
  "ok": true
}
````
````
$ curl --url vmpooler-dev/vm/gj3hpatv4qf9kfn
{
  "ok": true,
  "gj3hpatv4qf9kfn": {
    "template": "centos-6-x86_64",
    "lifetime": 2,
    "running": 0.01,
  }
}
````

PR #92, opened 2015-04-10
Redirect / to /dashboard

sschneid:
This PR also introduces a new `Vmpooler::Dashboard` library which contains all of the Sinatra routes to dashboards (currently just the one).

### DIFF
--- a/spec/vmpooler/api_spec.rb
+++ b/spec/vmpooler/api_spec.rb
@@ -11,13 +11,20 @@ describe Vmpooler::API do
   describe 'Dashboard' do
 
     context '/' do
+      before { get '/' }
+
+      it { expect(last_response.status).to eq(302) }
+      it { expect(last_response.location).to eq('http://example.org/dashboard/') }
+    end
+
+    context '/dashboard/' do
       let(:config) { {
           config: {'site_name' => 'test pooler'}
       } }
 
       before do
         $config = config
-        get '/'
+        get '/dashboard/'
       end
 
       it { expect(last_response).to be_ok }
--- a/lib/vmpooler/api.rb
+++ b/lib/vmpooler/api.rb
@@ -17,11 +17,19 @@ module Vmpooler
     end
 
     get '/' do
-      erb :dashboard, locals: {
-        site_name: $config[:config]['site_name'] || '<b>vmpooler</b>'
-      }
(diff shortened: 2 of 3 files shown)


spec/vmpooler/api_spec.rb
```diff
@@ -11,13 +11,20 @@ def app()
   describe 'Dashboard' do
 
     context '/' do
+      before { get '/' }
+
+      it { expect(last_response.status).to eq(302) }
+      it { expect(last_response.location).to match(/dashboard/) }
```
colinPL: This will regex match "dashboard" anywhere in the location. Do you want this or `eq('/dashboard/')` to verify the location is exactly just "/dashboard/"?

sschneid: It's a call to `/`, which should redirect to `dashboard`, so I was attempting to make sure that was set in the `location` headers.

````
$ curl -v --url vmpooler-dev/
````
````
* Hostname was NOT found in DNS cache
*   Trying 10.32.112.37...
* Connected to vmpooler-dev (10.32.112.37) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.37.1
> Host: vmpooler-dev
> Accept: */*
>
< HTTP/1.1 302 Found
* Server nginx is not blacklisted
< Server: nginx
< Date: Fri, 10 Apr 2015 17:01:19 GMT
< Content-Type: text/html;charset=utf-8
< Content-Length: 0
< Connection: keep-alive
< Location: http://vmpooler-dev/dashboard/
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
<
* Connection #0 to host vmpooler-dev left intact
````

Is that not sane?

Comment (colinPL) [inline spec/vmpooler/api_spec.rb]: I changed it to `eq` just to see. Here is the output:

```
 1) Vmpooler::API Dashboard / should eq "/dashboard/"
     Failure/Error: it { expect(last_response.location).to eq('/dashboard/') }
       
       expected: "/dashboard/"
            got: "http://example.org/dashboard/"
```

sschneid:
Updated spec to explicitly expect a redirect to `http://example.org/dashboard/`.

PR #94, opened 2015-04-15
(QENG-2208) Move Sinatra Helpers to own file

colinPL:
This moves the inline Helpers contained in V1.rb to their own file:
helpers.rb. In making this change, any API.settings call was removed
from the helper method itself and passed through from V1.

This also adds tests for hostname shortener and validate date string.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -5,133 +5,7 @@ module Vmpooler
     api_prefix  = "/api/v#{api_version}"
 
     helpers do
-      def get_capacity_metrics()
-        capacity = {
-          current: 0,
-          total: 0,
-          percent: 0
-        }
-
-        Vmpooler::API.settings.config[:pools].each do |pool|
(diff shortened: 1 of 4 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -144,8 +18,8 @@ def validate_date_str(date_str)
         }
       }
 
-      result[:capacity] = get_capacity_metrics()
-      result[:queue] = get_queue_metrics()
+      result[:capacity] = get_capacity_metrics(Vmpooler::API.settings.config[:pools], Vmpooler::API.settings.redis)
+      result[:queue] = get_queue_metrics(Vmpooler::API.settings.config[:pools], Vmpooler::API.settings.redis)
       result[:clone] = get_task_metrics('clone', Date.today.to_s)
       result[:boot] = get_task_metrics('boot', Date.today.to_s)
```
sschneid: These `get_task_metrics` calls need to be passed a Redis object.

sschneid:
:+1:

PR #95, opened 2015-04-15
(QENG-2208) Add more helper tests

colinPL:
Add tests for more Helper methods. This moves line coverage to 59%,
leaving only get_task_metrics remaining.

### DIFF
--- a/spec/vmpooler/api/helpers_spec.rb
+++ b/spec/vmpooler/api/helpers_spec.rb
@@ -37,4 +37,119 @@ describe Vmpooler::API::Helpers do
     end
   end
 
+  describe '#mean' do
+    [
+        [[1, 2, 3, 4], 2.5],
+        [[1], 1],
+        [[nil], 0],
+        [[], 0]
+    ].each do |list, expected|
+      it "returns #{expected.inspect} for #{list.inspect}" do
+        expect(subject.mean(list)).to eq expected
+      end
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


rji:
The patch looks good to me, and the tests are passing. Before I hit merge though, do we want a ticket number from a private JIRA project in the commit message?

waynr:
@rji we occasionally get private JIRA ticket references on other public projects:
https://github.com/puppetlabs/puppet/pull/3744

rji:
@waynr - That's a poor example and that PR shouldn't have been merged without modifying the subject of the commit message. Prior to a release, the Platform team runs a script to ensure all commit messages conform to their standard and either use `PUP`, `(maint)`, or several other keywords. If it doesn't fall within the public JIRA project, they require the commit message to be modified.

waynr:
There is utility value in having the ticket reference somewhere in the commit, in my opinion.

For puppet it is possible to use a public JIRA project to indirectly reference the removed JIRA ticket reference.

For vmpooler, there is no public JIRA project I am aware of that we can use as an alternative.

It's not clear to me what the value is in removing the non-public JIRA reference.

waynr:
Just to be clear, I am not suggesting that we always allow non-public JIRA references--it may be the case that we should actually create a public JIRA project for vmpooler that we can reference in commit messages. Just saying there is value in having some kind of reference to our ticketing system when inspecting past commits.

PR #96, opened 2015-04-16
Add basic [LDAP] authentication and /token routes

sschneid:
For review while I write spec tests and update docs (both needed before this should be merged).

This is deployed to and has been tested on vmpooler-dev.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -182,9 +182,11 @@ module Vmpooler
 
       result[params[:token]] = Vmpooler::API.settings.redis.hgetall('vmpooler__token__' + params[:token])
 
-      if Vmpooler::API.settings.config[:auth] and not result[params[:token]].nil?
+      if Vmpooler::API.settings.config[:auth] and result[params[:token]]['timestamp']
         status(200)
         result['ok'] = true
+      else
+        result.delete(params[:token])
       end
 
       JSON.pretty_generate(result)
@@ -197,11 +199,9 @@ module Vmpooler
 
       Vmpooler::API.settings.config[:auth] ? status(401) : status(404)
 
-      if Vmpooler::API.settings.config[:auth] and Vmpooler::API.settings.redis.exists('vmpooler__token__' + params[:token])
+      if Vmpooler::API.settings.config[:auth] and Vmpooler::API.settings.redis.del('vmpooler__token__' + params[:token])
         status(200)
         result['ok'] = true
-
-        Vmpooler::API.settings.redis.del('vmpooler__token__' + params[:token])
       end
 
       JSON.pretty_generate(result)
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -180,11 +180,11 @@ module Vmpooler
 
       Vmpooler::API.settings.config[:auth] ? status(401) : status(404)
 
-      if Vmpooler::API.settings.config[:auth] and Vmpooler::API.settings.redis.exists('vmpooler__token__' + params[:token])
+      result[params[:token]] = Vmpooler::API.settings.redis.hgetall('vmpooler__token__' + params[:token])
+
+      if Vmpooler::API.settings.config[:auth] and not result[params[:token]].nil?
         status(200)
         result['ok'] = true
-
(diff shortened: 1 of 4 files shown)


sschneid:
Basic usage testing can be seen at https://gist.github.com/sschneid/2914accbb9f5cc1ed314.

lib/vmpooler/api/v1.rb
```diff
@@ -173,6 +173,72 @@ class V1 < Sinatra::Base
       JSON.pretty_generate(result)
     end
 
+    get "#{api_prefix}/token/:token/?" do
+      content_type :json
+
+      result = {}
+
+      Vmpooler::API.settings.config[:auth] ? status(401) : status(404)
```
colinPL: Regardless the client receives a `40x` response, yes? How does it continue to the remainder of this route?

lib/vmpooler/api/v1.rb
```diff
@@ -173,6 +173,72 @@ class V1 < Sinatra::Base
...
+      result['ok'] = false
```
colinPL: Can this be moved up to the initial `result` statement? `result = {'ok' => false}`

lib/vmpooler/api/v1.rb
```diff
@@ -173,6 +173,72 @@ class V1 < Sinatra::Base
...
+
+      if Vmpooler::API.settings.config[:auth] and Vmpooler::API.settings.redis.exists('vmpooler__token__' + params[:token])
+        status(200)
+        result['ok'] = true
+
+        result[params[:token]] = Vmpooler::API.settings.redis.hgetall('vmpooler__token__' + params[:token])
```
colinPL: This should go before `status` and `result['ok']`. It is possible the `hgetall` call could fail. Or does Sinatra catch failures and always return a `5xx` in that case?

sschneid: If for some reason the key is deleted between the `exists` and the `hgetall`, Redis will return a `nil`.

````json
{
  "ok": true,
  "hwwzouwrg5a0k6ezk4lyy1ma4zzvtwg1": {

  }
}
````

Comment (colinPL) [inline lib/vmpooler/api/v1.rb]: Can the `exists` call be skipped? This would change the `404` assumption to mean does not exist **or** has no content.

``` ruby
if config[:auth]
  token_details = redis.hgetall(...)
end

unless token_details.nil?
  status(200)
  result['ok'] = true
end
```

sschneid: That make a lot more sense -- thanks.

lib/vmpooler/api/v1.rb
```diff
@@ -173,6 +173,69 @@ class V1 < Sinatra::Base
...
+        if authenticate(
+          Vmpooler::API.settings.config[:auth],
+          jdata['username'].to_s,
+          jdata['password'].to_s
+        )
+          status(200)
+          result['ok'] = true
+
+          o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
+          result['token'] = o[rand(25)] + (0...31).map { o[rand(o.length)] }.join
```
colinPL: This should be a helper, so that it can be reused in other routes later.

lib/vmpooler/api/v1.rb
```diff
@@ -173,6 +173,69 @@ class V1 < Sinatra::Base
...
+
+      result = { 'ok' => false }
+
+      Vmpooler::API.settings.config[:auth] ? status(401) : status(404)
+
+      if Vmpooler::API.settings.config[:auth] and Vmpooler::API.settings.redis.exists('vmpooler__token__' + params[:token])
+        status(200)
+        result['ok'] = true
+
+        Vmpooler::API.settings.redis.del('vmpooler__token__' + params[:token])
```
colinPL: `del` returns the number of keys deleted, 0 meaning the key did not exist. This can be setup similar to the `hgetall` comment above.

``` ruby
if config[:auth]
  del_count = redis.del(....)
end

if del_count > 1
  status(200)
  ....
end
```

lib/vmpooler/api/v1.rb
sschneid: I completely agree with this sentiment -- should I move app-wide methods like this into helpers in this PR, or wait?

colinPL: Let's do that in a separate ticket.

PR #97, opened 2015-04-22
Add basic HTTP authentication and /token routes

sschneid:


lib/vmpooler/api/helpers.rb
```diff
@@ -4,6 +4,60 @@ class API
...
+
+      def authenticate(auth, username_str, password_str)
+        case auth['provider']
+          when 'ldap'
+            require 'rubygems'
+            require 'net/ldap'
+
+            ldap = Net::LDAP.new(
+              :host => auth[:ldap]['host'],
+              :port => auth[:ldap]['port'] || 389,
```
rji: Since we default to TLS below, we probably want the default LDAP port to be 636

lib/vmpooler/api/helpers.rb
```diff
@@ -4,6 +4,60 @@ class API
...
+              :encryption => {
+                :method => :start_tls,
+                :tls_options => { :ssl_version => 'TLSv1' }
+              },
```
rji: We might want to consider making LDAP over SSL optional, although I can't think of any sane person wanting plaintext LDAP these days :smile:

rji:
Could this PR also update the docs and the example vmpooler config file?

lib/vmpooler/api/helpers.rb
sschneid: TLS usually happens over the non-SSL port.  :)

rji: According to the [OpenLDAP docs](http://www.openldap.org/doc/admin24/runningslapd.html), the use of `ldaps://` will default to 636. But if we make TLS optional in `authenticate()`, then my comments here can simply be ignored.

sschneid: Yeah, it's usually one or the other... straight SSL communication over :636 or start_tls encryption over :389.

In any case this could be more abstracted so that whoever is running the service can choose how they want the vmpooler app to communicate with LDAP.  As-is I'm kind of forcing the second option (start_tls over :389) or double-encryption (start_tls over :636).  :/

rji: That's a good point. In that case, I'm okay with defaulting to 389 here.

colinPL:
:+1: 

I think docs and examples can wait for a future PR. This was aimed at getting the functionality in place.

rji:
@colinPL New functionality requires new/updated documentation, which is very much related to this PR. Just like you wouldn't add new functionality without tests, I can't imagine adding new functionality without docs. And I still have an outstanding comment above about LDAP encryption being configurable.

rji:
@colinPL - that last comment was a bit premature; I talked to @sschneid and realized these tokens aren't being applied to our endpoints yet, so there isn't anything to document (and hence your comments above). That work is scheduled for another ticket in the future. So I'm also :+1: on this new auth functionality.

PR #103, opened 2015-05-13
Auto-expire summary index keys

sschneid:


sschneid:
This is breaking spec.  :-1:

sschneid:
Fixed spec.

lib/vmpooler/pool_manager.rb
```diff
@@ -66,6 +66,9 @@ def move_pending_vm_to_ready(vm, pool, host)
         $redis.smove('vmpooler__pending__' + pool, 'vmpooler__ready__' + pool, vm)
         $redis.hset('vmpooler__boot__' + Date.today.to_s, pool + ':' + vm, finish)
 
+        # Auto-expire summary index
+        $redis.expire('vmpooler__boot__' + Date.today.to_s, ($config[:redis]['data_ttl'].to_i * 60 * 60))
```
colinPL: Can this be moved to its own method and referenced below? This would ensure the ttl conversion calculation is in one location.

sschneid: ...yes.

PR #105, opened 2015-06-02
(QENG-2518) Tag-filtering

sschneid:
This PR adds regular expression-based tag-filtering functionality.  Given this configuration stanza:

````yaml
:tagfilter:
  url: '(.*)\/'
````

...any 'url' tag will be truncated after the first matching '/' character.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -542,8 +542,13 @@ module Vmpooler
                 backend.hset('vmpooler__vm__' + params[:hostname], param, arg)
               when 'tags'
                 arg.keys.each do |tag|
-                    backend.hset('vmpooler__vm__' + params[:hostname], 'tag:' + tag, arg[tag])
-                    backend.hset('vmpooler__tag__' + Date.today.to_s, params[:hostname] + ':' + tag, arg[tag])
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by sschneid** [comment]
Don't merge this yet, I think I want to match on what to keep rather than what to throw away.

File: lib/vmpooler/api/v1.rb [hunk 1]
```diff
@@ -542,8 +542,13 @@ def need_token!
                 backend.hset('vmpooler__vm__' + params[:hostname], param, arg)
               when 'tags'
                 arg.keys.each do |tag|
-                    backend.hset('vmpooler__vm__' + params[:hostname], 'tag:' + tag, arg[tag])
-                    backend.hset('vmpooler__tag__' + Date.today.to_s, params[:hostname] + ':' + tag, arg[tag])
+                  if Vmpooler::API.settings.config[:tagfilter] and Vmpooler::API.settings.config[:tagfilter][tag]
+                    filter = Regexp.new(Vmpooler::API.settings.config[:tagfilter][tag])
```
colinPL: I am not sure the expense of creating new Regexp objects. This should be moved to a hash that is either updated when new filters are found or at app initialization.

```
regexes[tag] ||= Regexp.new(...)
arg[tag] = arg[tag].gsub(regexes[tag], '')
```

PR #106, opened 2015-06-03
Only filter regex matches

sschneid:
and a spec test for it.

Previously using the example shown in vmpooler.yaml.example was failing
to tag strings WITHOUT a '/' in them.


rji:
:+1:

PR #107, opened 2015-06-04
API /summary rework

sschneid:
* moves the majority of summary-generation to helper methods
* implements `/summary/:route/:key` dynamic routes, eg.
  * `/summary/tag`
  * `/summary/tag/created_by`
  * `/summary/tag/created_by?from=2015-06-01`
  * `/summary/boot`
  * `/summary/clone`
  * `/summary/clone/duration?from=2015-05-29&to=2015-06-01`
  * ...and so on

PR #108, opened 2015-06-04
API summary rework

sschneid:
Bug in #107 squashed; resubmitting for review.

* moves the majority of summary-generation to helper methods
* implements `/summary/:route/:key` dynamic routes, eg.
  * `/summary/tag`
  * `/summary/tag/created_by`
  * `/summary/tag/created_by?from=2015-06-01`
  * `/summary/boot`
  * `/summary/clone`
  * `/summary/clone/duration?from=2015-05-29&to=2015-06-01`
  * ...and so on

### DIFF
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -147,7 +147,9 @@ module Vmpooler
         queue
       end
 
-      def get_tag_metrics(backend, date_str)
+      def get_tag_metrics(backend, date_str, opts = {})
+        opts = {:only => false}.merge(opts)
+
         tags = {}
 
         backend.hgetall('vmpooler__tag__' + date_str).each do |key, value|
@@ -156,6 +158,10 @@ module Vmpooler
 
           if key =~ /\:/
             hostname, tag = key.split(':', 2)
+          end
+
+          if opts[:only]
+            next unless tag == opts[:only]
           end
 
           tags[tag] ||= {}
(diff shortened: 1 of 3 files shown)


lib/vmpooler/api/helpers.rb
```diff
@@ -158,6 +160,10 @@ def get_tag_metrics(backend, date_str)
             hostname, tag = key.split(':', 2)
           end
 
+          if opts[:only]
+            next unless tag == opts[:only]
```
colinPL: Above, `:only` is a boolean (`:only => false`) and here it is being compared to a string (`tag = 'unknown').

sschneid: So... it fails!  :)

lib/vmpooler/api/v1.rb
```diff
@@ -121,90 +88,75 @@ def need_token!
...
-        end
+      case params[:route]
+        when 'boot'
+          result = get_task_summary(backend, 'boot', from_date, to_date, :bypool => true, :only => params[:key])
+        when 'clone'
+          result = get_task_summary(backend, 'clone', from_date, to_date, :bypool => true, :only => params[:key])
+        when 'tag'
+          result = get_tag_summary(backend, from_date, to_date, :only => params[:key])
+        else
+          halt 404, JSON.pretty_generate({ 'ok' => false })
```
colinPL: This should probably be a `400 Bad Request` error. The URL requested does exist, but the parameter is invalid.

sschneid: It's an invalid (not found) route, though -- eg. `/summary/something` is not found.  This matches the behavior of trying to query a nonexistant VM:
````
$ curl -v --url vmpooler/vm/foo
````
````
< HTTP/1.1 404 Not Found
* Server nginx is not blacklisted
< Server: nginx
< Date: Thu, 04 Jun 2015 20:32:06 GMT
< Content-Type: application/json
< Content-Length: 17
< Connection: keep-alive
< X-Content-Type-Options: nosniff
<
{
  "ok": false
}
````

colinPL: Makes sense. I thought this was a query param. `/summary?route=something` vs `/summary/something`.

colinPL:
:+1:

rji:
:+1:

PR #110, opened 2015-06-30
(QENG-2388) Tagging restrictions

sschneid:
- Allow for configurable 'allowed_tags' tag restrictions
- Skip empty tags

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -476,6 +476,10 @@ module Vmpooler
               unless arg.is_a?(Hash)
                 failure = true
               end
+
+              if config['allowed_tags']
+                failure = true if not (arg.keys - config['allowed_tags']).empty?
+              end
             else
               failure = true
           end
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -497,17 +497,8 @@ module Vmpooler
 
                 backend.hset('vmpooler__vm__' + params[:hostname], param, arg)
               when 'tags'
-                arg.keys.each do |tag|
-                  if Vmpooler::API.settings.config[:tagfilter] and Vmpooler::API.settings.config[:tagfilter][tag]
-                    filter = Vmpooler::API.settings.config[:tagfilter][tag]
-                    arg[tag] = arg[tag].match(filter).captures.join if arg[tag].match(filter)
-                  end
-
-                  next if arg[tag].nil? or arg[tag].empty?
-
-                  backend.hset('vmpooler__vm__' + params[:hostname], 'tag:' + tag, arg[tag])
-                  backend.hset('vmpooler__tag__' + Date.today.to_s, params[:hostname] + ':' + tag, arg[tag])
-                end
+                filter_tags(arg)
+                export_tags(backend, params[:hostname], arg)
             end
           end
 
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -499,6 +499,8 @@ module Vmpooler
                     arg[tag] = arg[tag].match(filter).captures.join if arg[tag].match(filter)
                   end
 
+                  next if arg[tag].nil? or arg[tag].empty?
+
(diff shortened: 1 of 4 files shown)


spec/vmpooler/api/v1_spec.rb
```diff
@@ -332,6 +341,23 @@ def app()
...
         end
 
+      context '(allowed_tags configured)' do
+        let(:config) { {
+          config: {
+            'allowed_tags' => ['created_by', 'project', 'url']
+          }
+        } }
+
+        it 'fails if restricted tag is specified' do
```
waynr: nit: The wording here is a little confusing, when I read it I thought that it implied there is some way to explicitly name restricted tags when it's actually just preventing the use of tags other than what are explicitly allowed.

sschneid: Excellent point, the wording *is* confusing.  I'll adjust.

lib/vmpooler/api/v1.rb
```diff
@@ -476,6 +476,14 @@ def need_token!
               unless arg.is_a?(Hash)
                 failure = true
               end
+
+              if config['allowed_tags']
+                arg.keys.each do |tag|
+                  unless config['allowed_tags'].include? tag
+                    failure = true
+                  end
```
rick: This loop could be avoided by subtracting `arg.keys` from `config['allowed_tags']` and checking for emptiness:

```
>> a = [1, 2, 3, 4, 5]
=> [1, 2, 3, 4, 5]
>> b = [ 2, 3, 4, 5, 6]
=> [2, 3, 4, 5, 6]
>> a - b
=> [1]
>> b - a
=> [6]
```

lib/vmpooler/api/v1.rb
```diff
@@ -499,6 +507,8 @@ def need_token!
                     arg[tag] = arg[tag].match(filter).captures.join if arg[tag].match(filter)
                   end
```
rick: While it's nominally outside the changes in the PR, it tends to be confusing (and often brittle) to modify a hash while it's being iterated.

I would see this block of code as being a good candidate for extraction to a method with an intention-revealing name, perhaps something like:

``` ruby
  when 'tags'
    arg = filter_tags(arg)
    export_tags(params[:hostname], arg)
```

with

``` ruby
  def filter_tags(tags = {})
    return unless Vmpooler::API.settings.config[:tagfilter]
    result = {}
    tags.each_pair do |tag, value|
      next unless filter = Vmpooler::API.settings.config[:tagfilter][tag]
      result[tag] = value.match(filter).captures.join if value.match(filter)
    end
  end

  def export_tags(hostname, tags)
    tags.each_pair do |tag, value|
      backend.hset('vmpooler__vm__' + hostname, 'tag:' + tag, value)
      backend.hset('vmpooler__tag__' + Date.today.to_s, hostname + ':' + tag, value)
  end
```

waynr: +1 on delegating to methods with intention-revealing names.

Also, I'm not sure what `backend` refers to here as the word is pretty generic and could refer to any number of things. It could just be that I'm not familiar enough with this code, if I had to guess though I'd say it refers to the redis backend? Is that something that is presumed to be swappable with other backends like postgres?

lib/vmpooler/api/v1.rb
sschneid: Updated.

lib/vmpooler/api/v1.rb
sschneid: > if I had to guess though I'd say it refers to the redis backend? Is that something that is presumed to be swappable with other backends like postgres?

Yes and yes (in the future).

sschneid: Updated in 5868469.

rick:
:+1: on all the things I commented on.

sschneid:
This branch is deployed on vmpooler-dev with the following config:

````json
  allowed_tags:
    - 'beaker_version'
    - 'created_by'
    - 'department'
    - 'jenkins_build_url'
    - 'project'
````

waynr:
:see_no_evil: But I don't feel comfortable merging yet since this is my first vmpooler review.

rji:
:+1:

PR #111, opened 2015-07-06
(QENG-2246) Add Travis CI

colinPL:
This commit adds a Rakefile, Gemfile, and Travis CI config file.

The Rakefile contains two tasks: test and junit. Both targets execute the spec tests, with the difference being how the results are presented.

The Gemfile defines dependencies and a test group. To install just the gems needed for running the application, execute:
  * `bundle install --without test`

The new .travis.yml file is for Travis CI and targets the major ruby versions.


sschneid:
:+1: looks good to me!

PR #117, opened 2015-07-14
(QENG-2636) Host snapshots

sschneid:
This PR introduces host (VM) snapshotting support.  Example workflow:

1) Request a VM
````
$ curl -d '{"debian-7-x86_64":"1"}' --url vmpooler-dev/vm
{
  "ok": true,
  "debian-7-x86_64": {
    "ok": true,
    "hostname": "cxd6n45t1pd5s4j"
  },
}
````

2) Take a snapshot
````
This command will return a snapshot ID ("fjgd9l9okym8647gmjnznky5sklji0r3")

$ curl -d --url vmpooler-dev/vm/cxd6n45t1pd5s4j/snapshot
{
  "ok": true,
  "cxd6n45t1pd5s4j": {
    "snapshot": "fjgd9l9okym8647gmjnznky5sklji0r3"
  }
}
````

3) The vmpooler pooler_manager component sees the snapshot-creation task, and snapshots the VM in the background
````
[2015-05-01 12:35:02] [ ] [snapshot_manager] 'cxd6n45t1pd5s4j' is being snapshotted
[2015-05-01 12:39:04] [+] [snapshot_manager] 'cxd6n45t1pd5s4j' snapshot created in 242.77 seconds
````

4) You can query the VM itself; the GET /vm/:hostname endpoint will show a snapshot once pool_manager has finished the snapshotting task
````
$ curl --url vmpooler-dev/vm/cxd6n45t1pd5s4j
{
  "ok": true,
  "cxd6n45t1pd5s4j": {
    "template": "debian-7-x86_64",
    "lifetime": 1,
    "running": 0.08,
    "snapshots": [
      "fjgd9l9okym8647gmjnznky5sklji0r3"
    ],
  }
}
````

4) Change something on the VM
````
$ ssh root@cxd6n45t1pd5s4j
root@cxd6n45t1pd5s4j's password:
Linux debian-7-64 3.2.0-4-amd64 #1 SMP Debian 3.2.65-1 x86_64 

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright. 

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Tue Mar 31 08:33:05 2015
root@cxd6n45t1pd5s4j:~# echo "This VM has been modified." >/etc/motd
root@cxd6n45t1pd5s4j:~# cat /etc/motd
This VM has been modified.
````

5) Revert the VM back to the snapshot we took earlier
````
$ curl -d --url vmpooler-dev/vm/cxd6n45t1pd5s4j/snapshot/fjgd9l9okym8647gmjnznky5sklji0r3
{
  "ok": true
}
````

6) pool_manager reverts the VM
````
[2015-05-01 12:41:52] [ ] [snapshot_manager] 'cxd6n45t1pd5s4j' is being reverted to snapshot 'fjgd9l9okym8647gmjnznky5sklji0r3'
[2015-05-01 12:41:58] [<] [snapshot_manager] 'cxd6n45t1pd5s4j' reverted to snapshot in 6.86 seconds
````

7) Verify that the changes in step 4 do not persist
````
$ ssh root@cxd6n45t1pd5s4j
root@cxd6n45t1pd5s4j's password:
Linux debian-7-64 3.2.0-4-amd64 #1 SMP Debian 3.2.65-1 x86_64 

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright. 

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Tue Mar 31 08:33:05 2015
root@cxd6n45t1pd5s4j:~# cat /etc/motd 

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright. 

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
````

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -413,6 +413,11 @@ module Vmpooler
             result[params[:hostname]]['tags'] ||= {}
             result[params[:hostname]]['tags'][$1] = rdata[key]
           end
+
+          if key.match('^snapshot\:(.+?)$')
+            result[params[:hostname]]['snapshots'] ||= []
+            result[params[:hostname]]['snapshots'].push($1)
+          end
         end
 
         if config['domain']
@@ -505,6 +510,47 @@ module Vmpooler
           status 200
           result['ok'] = true
         end
+      end
+
+      JSON.pretty_generate(result)
+    end
+
+    post "#{api_prefix}/vm/:hostname/snapshot/?" do
+      content_type :json
+
+      status 404
+      result = { 'ok' => false }
+
+      params[:hostname] = hostname_shorten(params[:hostname], config['domain'])
+
+      if backend.exists('vmpooler__vm__' + params[:hostname])
+        result[params[:hostname]] = {}
+
+        o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
+        result[params[:hostname]]['snapshot'] = o[rand(25)] + (0...31).map { o[rand(o.length)] }.join
+
+        backend.sadd('vmpooler__tasks__snapshot', params[:hostname] + ':' + result[params[:hostname]]['snapshot'])
+
+        status 202
+        result['ok'] = true
+      end
+
+      JSON.pretty_generate(result)
+    end
+
+    post "#{api_prefix}/vm/:hostname/snapshot/:snapshot/?" do
+      content_type :json
+
+      status 404
+      result = { 'ok' => false }
+
+      params[:hostname] = hostname_shorten(params[:hostname], config['domain'])
+
+      if backend.exists('vmpooler__vm__' + params[:hostname]) and backend.hget('vmpooler__vm__' + params[:hostname], 'snapshot:' + params[:snapshot])
+        backend.sadd('vmpooler__tasks__snapshot-revert', params[:hostname] + ':' + params[:snapshot])
+
+        status 202
+        result['ok'] = true
       end
 
       JSON.pretty_generate(result)
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -518,6 +518,8 @@ module Vmpooler
     post "#{api_prefix}/vm/:hostname/snapshot/?" do
       content_type :json
 
+      need_token! if Vmpooler::API.settings.config[:auth]
+
       status 404
       result = { 'ok' => false }
 
@@ -540,6 +542,8 @@ module Vmpooler
 
     post "#{api_prefix}/vm/:hostname/snapshot/:snapshot/?" do
       content_type :json
+
+      need_token! if Vmpooler::API.settings.config[:auth]
 
       status 404
       result = { 'ok' => false }
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -286,6 +286,100 @@ module Vmpooler
       end
     end
 
+    def create_vm_snapshot(vm, snapshot_name)
+      Thread.new do
+        _create_vm_snapshot(vm, snapshot_name)
+      end
+    end
+
+    def _create_vm_snapshot(vm, snapshot_name)
+      host = $vsphere['snapshot_manager'].find_vm(vm) ||
+             $vsphere['snapshot_manager'].find_vm_heavy(vm)[vm]
+
+      if (host) && ((! snapshot_name.nil?) || (! snapshot_name.empty?))
+        $logger.log('s', "[ ] [snapshot_manager] '#{vm}' is being snapshotted")
+
+        start = Time.now
+
+        host.CreateSnapshot_Task(
+          name: snapshot_name,
+          description: 'vmpooler',
+          memory: true,
+          quiesce: true
+        ).wait_for_completion
+
+        finish = '%.2f' % (Time.now - start)
+
+        $redis.hset('vmpooler__vm__' + vm, 'snapshot:' + snapshot_name, Time.now.to_s)
+
+        $logger.log('s', "[+] [snapshot_manager] '#{vm}' snapshot created in #{finish} seconds")
+      end
+    end
+
+    def revert_vm_snapshot(vm, snapshot_name)
+      Thread.new do
+        _revert_vm_snapshot(vm, snapshot_name)
+      end
+    end
+
+    def _revert_vm_snapshot(vm, snapshot_name)
+      host = $vsphere['snapshot_manager'].find_vm(vm) ||
+             $vsphere['snapshot_manager'].find_vm_heavy(vm)[vm]
+
+      if host
+        snapshot = $vsphere['snapshot_manager'].find_snapshot(host, snapshot_name)
+
+        if snapshot
+          $logger.log('s', "[ ] [snapshot_manager] '#{vm}' is being reverted to snapshot '#{snapshot_name}'")
+
+          start = Time.now
+
+          snapshot.RevertToSnapshot_Task.wait_for_completion
+
+          finish = '%.2f' % (Time.now - start)
+
+          $logger.log('s', "[<] [snapshot_manager] '#{vm}' reverted to snapshot in #{finish} seconds")
+        end
+      end
+    end
+
+    def check_snapshot_queue
+      $logger.log('d', "[*] [snapshot_manager] starting worker thread")
+
+      $vsphere['snapshot_manager'] ||= Vmpooler::VsphereHelper.new
+
+      $threads['snapshot_manager'] = Thread.new do
+        loop do
+          _check_snapshot_queue
+          sleep(5)
+        end
+      end
+    end
+
+    def _check_snapshot_queue
+      vm = $redis.spop('vmpooler__tasks__snapshot')
+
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by sschneid** [comment]
This feature still requires documentation to be written, but I wanted to push the code and rspec up to get an initial review.

**Comment by colinPL** [comment]
How many snapshots can a VM have at one time?

File: lib/vmpooler/api/v1.rb [hunk 1]
```diff
@@ -509,6 +514,47 @@ def need_token!
...
+      status 404
+      result = { 'ok' => false }
+
+      params[:hostname] = hostname_shorten(params[:hostname], config['domain'])
+
+      if backend.exists('vmpooler__vm__' + params[:hostname])
+        result[params[:hostname]] = {}
+
+        o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
+        result[params[:hostname]]['snapshot'] = o[rand(25)] + (0...31).map { o[rand(o.length)] }.join
```
colinPL: There isn't a random name method available?

sschneid:
> How many snapshots can a VM have at one time?

No limit defined by either vmpooler or vSphere.

lib/vmpooler/api/v1.rb
sschneid: > There isn't a random name method available?

Not currently.

lib/vmpooler/pool_manager.rb
```diff
@@ -292,6 +292,100 @@ def destroy_vm(vm, pool)
       end
     end
 
+    def create_vm_snapshot(vm, snapshot_name)
+      Thread.new do
+        _create_vm_snapshot(vm, snapshot_name)
+      end
+    end
+
+    def _create_vm_snapshot(vm, snapshot_name)
```
colinPL: This method should do some scrutiny of `snapshot_name`. Or is that left to `CreateSnapshot_Task`?

sschneid: What sort of scrutiny do you have in mind?  As-is the only method calling `create_vm_snapshot` is the API, which is generating the `snapshot_name` var right before calling this method.

lib/vmpooler/pool_manager.rb
```diff
@@ -443,6 +537,15 @@ def execute!
       $redis.set('vmpooler__tasks__clone', 0)
 
       loop do
+        if ! $threads['snapshot_manager']
```
colinPL: Can this be changed to `unless` or is the `!` needed to force a boolean?

lib/vmpooler/pool_manager.rb
```diff
@@ -443,6 +537,15 @@ def execute!
...
+          check_snapshot_queue
+        else
+          unless $threads['snapshot_manager'].alive?
```
colinPL: This should be merge to an `else-if`.

lib/vmpooler/api/v1.rb
rick: Would definitely say this pair of lines is a candidate for extracting to a method with an intention-revealing name.

``` ruby
  def random_snapshot_name
    o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
    o[rand(25)] + (0...31).map { o[rand(o.length)] }.join
  end

  # ...

    post "#{api_prefix}/vm/:hostname/snapshot/?" do
      content_type :json

      status 404
      result = { 'ok' => false }

      hostname = params[:hostname] = hostname_shorten(params[:hostname], config['domain'])

      if backend.exists("vmpooler__vm__#{hostname}")
        result[hostname] = {}
        result[hostname]['snapshot'] = random_snapshot_name
        backend.sadd('vmpooler__tasks__snapshot', hostname + ':' + result[hostname]['snapshot'])

        status 202
        result['ok'] = true
      end

      JSON.pretty_generate(result)
    end
```

lib/vmpooler/pool_manager.rb
sschneid: It was to match the `! $threads[pool['name']]` check at line 550 below.  I can modify both if you'd prefer `unless`.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -99,6 +99,14 @@ def find_pool(poolname)
       base
     end
 
+    def find_snapshot(vm, snapshotname)
+      if vm.snapshot
+        get_snapshot_list(vm.snapshot.rootSnapshotList, snapshotname)
+      else
+        raise "VM #{vm.name} has no snapshots to revert to"
```
colinPL: What about just returning an empty list instead of `raise`? Is not having a snapshot a reason to throw an error?

lib/vmpooler/vsphere_helper.rb
```diff
@@ -178,6 +186,20 @@ def get_base_vm_container_from(connection)
       )
     end
 
+    def get_snapshot_list(tree, snapshotname)
+      snapshot = nil
```
colinPL: The method name is `_list` - this should be initialized as an empty list `[]`.

lib/vmpooler/api/v1.rb
sschneid: I agree with the usefulness of a `create_a_random_string_of_N_characters` method.  Currenlty both `pool_manager` and `api/v1` have randomization code in them, and unfortunately there isn't currently a helper library which is shared throughout the app.  I can certainly refactor some code to accomplish this, but am unsure if it's out of the scope of the PR.

lib/vmpooler/pool_manager.rb
colinPL: Mostly a check that it is not `nil` or `empty?`.

lib/vmpooler/pool_manager.rb
colinPL: No - I would prefer them to match. This is fine for now.

lib/vmpooler/api/v1.rb
colinPL: I would ticket the creation of the method and work on it separately. It is out-of-scope of this PR, IMO.

rick: Agreed, defer for separate work.

spec/vmpooler/api/v1_spec.rb
```diff
@@ -432,6 +432,34 @@ def app()
...
+    describe 'POST /vm/:hostname/snapshot' do
+        it 'creates a snapshot' do
+          expect(redis).to receive(:sadd)
+
+          post "#{prefix}/vm/testhost/snapshot"
+
+          expect(last_response.header['Content-Type']).to eq('application/json')
+          expect(last_response.body).to include('"ok": true')
+          expect(last_response.body).to include('"testhost": {')
+          expect(last_response.body).to include('"snapshot": "')
```
colinPL: The above 3 lines can be altered to check for key existence if you parse the body as JSON:

```
json_response = JSON.parse(last_response.body)
expect(json_response['ok']).not_to be_nil
```

stahnma:
Random Question, will you have to be auth'ed to use snap shotting?

sschneid:
> Random Question, will you have to be auth'ed to use snap shotting?

That's actually something I started thinking about yesterday.  This PR doesn't require auth at the moment, but as snapshotting is a relatively "expensive" vSphere task, I think it probably should.

sschneid:
This is ready for re-review.  Addressed applicable comments above and included an additional commit to support requiring authentication tokens when using snapshots.

sschneid:
I've deployed this branch to `vmpooler-dev` for manual testing as well.

sschneid:
Some UX things I haven't figured out how to solve yet:

* no way for user to tell when snapshotting is complete other than to check `/api/v1/vm/:vm` until a `snapshots` section appears in the JSON response
* no way for user to tell when a snapshot revert task has completed

lib/vmpooler/api/v1.rb
```diff
@@ -413,6 +413,11 @@ def need_token!
             result[params[:hostname]]['tags'] ||= {}
             result[params[:hostname]]['tags'][$1] = rdata[key]
           end
+
+          if key.match('^snapshot\:(.+?)$')
```
colinPL: This could be an `elsif` for the above `if`.

lib/vmpooler/api/v1.rb
```diff
@@ -509,6 +514,51 @@ def need_token!
...
+      content_type :json
+
+      need_token! if Vmpooler::API.settings.config[:auth]
+
+      status 404
+      result = { 'ok' => false }
+
+      params[:hostname] = hostname_shorten(params[:hostname], config['domain'])
+
+      if backend.exists('vmpooler__vm__' + params[:hostname]) and backend.hget('vmpooler__vm__' + params[:hostname], 'snapshot:' + params[:snapshot])
```
colinPL: The `backend.exists` call can be ignored. `hget` will return `nil` if the key does not exist.

PR #119, opened 2015-07-28
(maint) Use expect_json helper method for determining JSON response status

sschneid:


spec/helpers.rb
```diff
@@ -0,0 +1,14 @@
+def expect_json(
+  ok: true,
```
colinPL: `:` should be `=`

spec/helpers.rb
```diff
@@ -0,0 +1,14 @@
...
+  http: 200
```
colinPL: `:` should be `=`

PR #120, opened 2015-07-28
(MAINT) Reduce redis Calls in API

colinPL:
The return values from most redis calls inform the caller of whether a
key or hash value exists. Several exists() calls can be removed in
favor of this approach.

Updated spec tests to account for a removal of exists() and ismember()
calls in API tests.


sschneid:
:+1: looks sane to me.

PR #124, opened 2015-08-19
Docs update

sschneid:


colinPL:
The `API.md` is a new doc, but the first heading is a h3 (`###`). Can this reset with the first heading being `#`?

sschneid:
> The API.md is a new doc, but the first heading is a h3 (###).

This was to match the typeface sizes in the original markdown.  I considered adding the other `#` header stuff in the original `README.md` but decided that it added clutter.

PR #127, opened 2015-08-21
(maint) Remove duplicate (nested) "ok" responses

sschneid:
As we approach an "official" v1.0.0 of the API I'd like to remove some old
nested "ok" responses.  These were left in as the Beaker vmpooler
hypervisor used them, but I long-ago patched that code and I think it's
time to deprecate these.


sschneid:
This is deployed on `vmpooler-dev` and my manual testing with Beaker indicates that everything continues to function identically.

stahnma:
LGTM.

PR #133, opened 2015-10-02
(MAINT) Remove Ping Check on Running VMs

colinPL:
Prior to this commit, a running VM could fail a ping check and be destroyed. This causes issues when network hiccups occur or the machine is performing a reboot.

A VM that is in a ready state will now be destroyed when handed back or it hits the lifetime TTL.


sschneid:
:+1:

PR #137, opened 2015-11-04
(QENG-2995) Display associated VMs in GET /token/:token endpoint

sschneid:
This is deployed on `vmpooler-dev`.

````
$ curl -k -X GET --url https://vmpooler-dev.delivery.puppetlabs.net/api/v1/token/cxig75je6dzjjinbg5oxmfjup63libs5
````
````json
{
  "ok": true,
  "cxig75je6dzjjinbg5oxmfjup63libs5": {
    "user": "sschneider",
    "created": "2015-11-03 14:56:57 -0800",
    "last": "2015-11-04 12:28:37 -0800",
    "vms": {
      "running": [
        "dqs4914g2wjyy5w",
        "hul7ib0ssr0f4o0"
      ]
    }
  }
}
````

### DIFF
--- a/API.md
+++ b/API.md
@@ -43,8 +43,7 @@ Enter host password for user 'jdoe':
 Get information about an existing token (including associated VMs).
 
 ```
-$ curl -u jdoe --url vmpooler.company.com/api/v1/token/utpg2i2xswor6h8ttjhu3d47z53yy47y
-Enter host password for user 'jdoe':
+$ curl --url vmpooler.company.com/api/v1/token/utpg2i2xswor6h8ttjhu3d47z53yy47y
 ```
 ```json
 {
--- a/API.md
+++ b/API.md
@@ -40,7 +40,7 @@ Enter host password for user 'jdoe':
 
 ##### GET /token/&lt;token&gt;
 
-Get information about an existing token.
+Get information about an existing token (including associated VMs).
 
 ```
 $ curl -u jdoe --url vmpooler.company.com/api/v1/token/utpg2i2xswor6h8ttjhu3d47z53yy47y
@@ -51,7 +51,14 @@ Enter host password for user 'jdoe':
   "ok": true,
   "utpg2i2xswor6h8ttjhu3d47z53yy47y": {
     "user": "jdoe",
-    "timestamp": "2015-04-28 19:17:47 -0700"
+    "created": "2015-04-28 19:17:47 -0700",
+    "last": "2015-11-04 12:28:37 -0700",
+    "vms": {
+      "running": [
+        "dqs4914g2wjyy5w",
+        "hul7ib0ssr0f4o0"
+      ]
+    }
   }
 }
 ```
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -209,6 +209,17 @@ module Vmpooler
 
         if not token.nil? and not token.empty?
           status 200
+
+          pools.each do |pool|
+            backend.smembers('vmpooler__running__' + pool['name']).each do |vm|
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by sschneid** [comment]
cc: @briancain @cowofevil @kevpl @stahnma  as this change may affect vmpooler clients.

**Comment by stahnma** [comment]
@sschneid could I send the token I have rather than do a username/pw auth?

**Comment by sschneid** [comment]
> could I send the token I have rather than do a username/pw auth?

It depends on if you think the information being returned needs some form of authentication or not.  Without requiring a username/password combination, the route could theoretically be brute-forced to discover existing tokens.

**Comment by stahnma** [comment]
Hmm. What if you always return 404 if the token isn't in the header or if the user isn't authed? Would that protect it? (No idea how difficult that is)

**Comment by briancain** [comment]
:+1: works for me. Shouldn't break anything I have currently! Really excited to see this get in.

**Comment by briancain** [comment]
Also for what it's worth, I think this endpoint already requires a login to query token status?

**Comment by sschneid** [comment]
@stahnma d74c9ff removed the authentication requirement for this endpoint.

**Comment by stahnma** [comment]
I'm not 100% sure I have the best ideas. It feels wrong to suggest something less secure. Do other want to weigh in? (I was trying to have a conversation more than suggest we follow my ideas.)

**Comment by briancain** [comment]
Personally I'd prefer the authentication for the endpoint considering the other two operations on tokens (GET and DELETE) also have it. Especially since tokens are often just stored on disk some where or could potentially be committed to Github in a .fog file.

**Comment by sschneid** [comment]
I have mixed feelings, too.  It's "less secure", but in the same vein, someone could check out a VM anonymously and attempt to change it's `lifetime` using brute-force tokens just as well as they could attack the `/token/:token` endpoint...

**Comment by sschneid** [comment]
FWIW, token-related endpoints would be:

- `GET /token' - requires username/password (returns list of all of  the user's tokens)
- `POST /token` - requires uername/password (generated new token)
- `GET /token/:token` - no auth (returns information about specified token)
- `DELETE /token/:token' - requires usename/password (deletes existing token)

**Comment by nicklewis** [comment]
Why not just make the tokens longer (and also generate them more securely...)? If the entropy of a token is greater than the entropy of the average LDAP password, password-protecting the `/token/:token` endpoint doesn't actually increase security.

**Comment by sschneid** [comment]
The token is 32 character and alphanumeric, so the total number of possible combinations would be 32^(26+10), or 63,340,286,662,973,277,706,162,286,946,811,886,609,896,461,828,096.

I guess that's pretty hard to brute-force.

**Comment by rick** [comment]
I don't know anything about how the tokens are generated, which would have a lot to do with how secure they are.  Presuming they're generated securely:

```
$ irb
>> Math.log(36**32, 2)
=> 165.437600046154
```

So `2^165` combinations, etc.

If one is in possession of a token, username/password auth shouldn't be required for listing things about the token. I could see a good case for saying the token could be deleted with only the token as well, but it doesn't hurt to have username/password auth there.

While I think there is some protection available in saying to require username/password in addition to the token for listing vm's, that doesn't seem like a standard practice, and it promotes storing not just tokens but username/password in more places where folks are using tokens. Seems like not the best thing.

sschneid:
> While I think there is some protection available in saying to require username/password in addition to the token for listing vm's, that doesn't seem like a standard practice, and it promotes storing not just tokens but username/password in more places where folks are using tokens. Seems like not the best thing.

I agree with all of this.

haus:
:+1:

nwolfe:
Perfect, @haus and I were just looking for this in the API yesterday!

> While I think there is some protection available in saying to require username/password in addition to the token for listing vm's, that doesn't seem like a standard practice, and it promotes storing not just tokens but username/password in more places where folks are using tokens. Seems like not the best thing.

:+1:

PR #138, opened 2015-11-05
(QENG-2807) Allow pool 'alias' names

sschneid:
The following pool configuration would allow a pool to be aliased in POST requests as '`centos-6-x86_64`', '`centos-6-amd64`', or '`centos-6-64`':

````yaml
- name: 'centos-6-x86_64'
  alias: [ 'centos-6-amd64', 'centos-6-64' ]
  template: 'templates/centos-6-x86_64'
  folder: 'vmpooler/centos-6-x86_64'
  datastore: 'instance1'
  size: 5
````

The `alias` configuration can be either a string or an array.

Note that even when requesting an alias, the pool's `name` is returned in the JSON response:

````
$ curl -d '{"centos-6-64":"1"}' --url vmpooler/api/v1/vm
````
````json
{
  "ok": true,
  "centos-6-x86_64": {
    "hostname": "cuna2qeahwlzji7"
  },
  "domain": "company.com"
}
````

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -28,6 +28,60 @@ module Vmpooler
       validate_token(backend)
     end
 
+    def alias_deref(hash)
+      newhash = {}
+
+      hash.each do |key, val|
+        if backend.exists('vmpooler__ready__' + key)
+          newhash[key] = val
+        else
+          if Vmpooler::API.settings.config[:alias][key]
+            newkey = Vmpooler::API.settings.config[:alias][key]
+            newhash[newkey] = val
+          end
+        end
+      end
+
+      newhash
+    end
+
+    def checkout_vm(template, result)
+      vm = backend.spop('vmpooler__ready__' + template)
+
+      unless vm.nil?
+        backend.sadd('vmpooler__running__' + template, vm)
+        backend.hset('vmpooler__active__' + template, vm, Time.now)
+        backend.hset('vmpooler__vm__' + vm, 'checkout', Time.now)
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by sschneid** [comment]
- This is deployed on `vmpooler-dev`
- This needs spec tests before being merged, but I figured I'd open the functional code up for review

**Comment by stahnma** [comment]
Did the JSON always return the pool name?

File: lib/vmpooler/api/v1.rb [hunk 1]
```diff
@@ -28,6 +28,60 @@ def need_token!
...
+            newkey = Vmpooler::API.settings.config[:alias][key]
+            newhash[newkey] = val
+          end
+        end
+      end
+
+      newhash
+    end
+
+    def checkout_vm(template, result)
```
stahnma: This is basically the same content that was here before, just in a different spot in the file, correct?

sschneid: Yeah, the [essentially] same code block was included twice for both `POST /vm` and `POST /vm/:template`, so I wrapped it up into it's own method (`checkout_vm`).

sschneid:
> Did the JSON always return the pool name?

Yup.

stahnma:
I like it.

sschneid:
Spec tests added.  As a result of writing spec tests, I discovered a bug in my original code which has also been patched.

stahnma:
:+1:

waynr:
:+1: but I'll wait for @sschneid to indicate he is ready for merge since I am not familiar with the production vmpooler deployment.

PR #142, opened 2015-12-14
[DO NOT MERGE](maint) Remove Some Thread.new Calls

colinPL:
Remove the creation of threads when mostly interacting with redis. This reduces the number of threads created during a spiking activity.

As a side effect, it is possible that a pool can take a long time checking the port availability of new hosts. This timeout will be made configurable in a future change.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -22,12 +22,6 @@ module Vmpooler
 
     # Check the state of a VM
     def check_pending_vm(vm, pool, timeout)
-      Thread.new do
-        _check_pending_vm(vm, pool, timeout)
-      end
-    end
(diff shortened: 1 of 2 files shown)


colinPL:
Do not merge!

stahnma:
@rick or @shermdog should we work to get this in or close it?

rick:
I'm in the "close it" camp. If it's a known issue, keep this for the historical record, but I'm not even sure (a) how out of date on current master it is, and (b) whether it's even the right implementation.

If @shermdog or you are more clear on the need for it, or whether it's the right approach, feel free to disregard me.

shermdog:
I'm going to close this for now.  We are going to have a larger conversation about thread usage in the near future.

PR #145, opened 2016-01-05
Updated YAML config variables in create_template_deltas.rb

frozenfoxx:
The symbol notation for the YAML variables "server," "username," and, "password," were not working for me.  When changing them to strings the script seems to work.


colinPL:
:+1:

PR #148, opened 2016-01-14
Update license copyright

sschneid:
Happy New Year!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
    vmpooler
 
-   Copyright (C) 2015 Puppet Labs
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
    vmpooler
 
-   Copyright (C) 2015 Puppet Labs
+   Copyright (C) 2016 Puppet Labs
```
stahnma: Shouldn't it be 2014-2016?

sschneid: Updated.

PR #149, opened 2016-01-14
Added prefix parameter to the vmpooler configuration

frozenfoxx:
I've found it useful to be able to specify a prefix to all created VMs.  This can help with defining autosign policy for puppet ("sign $prefix-<random string>") as well as VM organization.  This adds an optional parameter that lets the user do exactly this.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -189,7 +189,7 @@ module Vmpooler
 
         # Generate a randomized hostname
         o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
-        vm['hostname'] = o[rand(25)] + (0...14).map { o[rand(o.length)] }.join
+        vm['hostname'] = $config[:config]['prefix'] + o[rand(25)] + (0...14).map { o[rand(o.length)] }.join
 
         # Add VM to Redis inventory ('pending' pool)
         $redis.sadd('vmpooler__pending__' + vm['template'], vm['hostname'])
--- a/vmpooler.yaml.example
+++ b/vmpooler.yaml.example
@@ -163,7 +163,9 @@
 #     If set, returns a top-level 'domain' JSON key in POST requests
 #
 #   - prefix
-#     If set, prefixes all created VMs with this string.
+#     If set, prefixes all created VMs with this string.  This should include
+#     a separator.
+#     (optional; default: '')
 
 # Example:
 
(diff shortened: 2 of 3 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -189,7 +189,7 @@ def clone_vm(template, folder, datastore, target)
 
         # Generate a randomized hostname
         o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
-        vm['hostname'] = o[rand(25)] + (0...14).map { o[rand(o.length)] }.join
+        vm['hostname'] = $config[:config]['prefix'] + o[rand(25)] + (0...14).map { o[rand(o.length)] }.join
```
colinPL: Is the separator between `prefix` and random name expected to be contained within the `prefix` itself?

vmpooler.yaml.example
```diff
@@ -161,6 +161,9 @@
 #
 #   - domain
 #     If set, returns a top-level 'domain' JSON key in POST requests
+#
+#   - prefix
+#     If set, prefixes all created VMs with this string.
```
colinPL: Please add `(optional)` to note that this is an optional setting.

Please add a comment indicating this setting should include the separator between `prefix` and the rest of the hostname.

vmpooler.yaml.example
```diff
@@ -176,6 +179,7 @@
     - 'created_by'
     - 'project'
   domain: 'company.com'
+  prefix: ''
```
colinPL: The value in the example configuration file should be populated with an example value, similar to `domain` on the line above. Please include an example prefix as this setting is intended to be used.

lib/vmpooler/pool_manager.rb
frozenfoxx: Yes, I figure different users might have different conventions (underscore, dash, colon, etc) and didn't want to make the assumption for them.

vmpooler.yaml.example
frozenfoxx: Gotcha, I'll get that added.

vmpooler.yaml.example
frozenfoxx: Makes sense, sorry I didn't think about that.  Adding one now.

colinPL:
:+1:

sschneid:
:+1:

PR #150, opened 2016-05-10
(WIP) (RE-7014) Add support for statsd

shermdog:
They way we were using graphite was incorrect for the type of data we were sending it.  statsd is the appropriate mechanism for our needs.
statsd and graphite are mutually exclusive and configuring statsd will take precendence over Graphite.  Example of configuration in vmpooler.yaml.example

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -1,12 +1,17 @@
 module Vmpooler
   class PoolManager
-    def initialize(config, logger, redis, graphite=nil)
+    def initialize(config, logger, redis, graphite=nil, statsd=nil)
       $config = config
 
       # Load logger library
       $logger = logger
 
-      unless graphite.nil?
+      # statsd and graphite are mutex in the context of vmpooler
+      unless statsd.nil?
+        $statsd = statsd
+      end
+
+      unless graphite.nil? || !statsd.nil?
         $graphite = graphite
       end
 
@@ -258,6 +263,7 @@ module Vmpooler
         $redis.decr('vmpooler__tasks__clone')
 
         begin
+          $statsd.timing($config[:statsd]['prefix'] + ".clone.#{vm['template']}", finish) if defined? $statsd
           $graphite.log($config[:graphite]['prefix'] + ".clone.#{vm['template']}", finish) if defined? $graphite
         rescue
         end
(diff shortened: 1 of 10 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -1,12 +1,17 @@
...
-    def initialize(config, logger, redis, graphite=nil)
+    def initialize(config, logger, redis, graphite=nil, statsd=nil)
       $config = config
 
       # Load logger library
       $logger = logger
 
-      unless graphite.nil?
+      # statsd and graphite are mutex in the context of vmpooler
+      unless statsd.nil?
```
mckern: I've got more to say about this in another comment I'm working on, but `nil` is already falsey, so you can just use the status of something that *might* be `nil` instead of explicitly checking for it.

```diff
 - unless graphite.nil?
 + if graphite
```

shermdog: Word - I generally just follow what ever pattern was done before, but I'm by no means opposed to cleaning this up as we go :)

mckern: I prefer the latter approach if we can try to match general style.

lib/vmpooler/pool_manager.rb
```diff
@@ -258,6 +263,7 @@ def clone_vm(template, folder, datastore, target)
         $redis.decr('vmpooler__tasks__clone')
 
         begin
+          $statsd.timing($config[:statsd]['prefix'] + ".clone.#{vm['template']}", finish) if defined? $statsd
           $graphite.log($config[:graphite]['prefix'] + ".clone.#{vm['template']}", finish) if defined? $graphite
```
mckern: It looks like you're checking on `$statsd` and `$graphite` a lot. Consider letting the initializer define both of those values (and letting one fall through to `nil`, and then wrapping those into a general method (this is by no means complete but if you're interested then I can help you flesh it out):

```ruby
# @return [SomeKindOfObject, nil] an object representing the defined processing system for metric collection
def metric_collector
  $statsd || $graphite
end
```

lib/vmpooler/statsd.rb
```diff
@@ -0,0 +1,12 @@
+require 'rubygems' unless defined?(Gem)
+
+module Vmpooler
+  class Statsd
+    def initialize(
+      s = 'statsd',
+      port = 8125
+    )
```
mckern: A nitpick and a must-do:

##### Imperative
Don't declare single letter variable names. Ever. Maybe in iterators, but definitely never in initializers.

##### Nitpick
The newlines in the required arguments don't help any. It can all be on one line.

lib/vmpooler/statsd.rb
```diff
@@ -0,0 +1,12 @@
+require 'rubygems' unless defined?(Gem)
```
mckern: This can go away. `rubygems` is defined and included by default in Ruby 1.9.1 and up. Throw this line away.

rick:
Random aside, if we are using statsd, it's worth knowing about brubeck: 

http://githubengineering.com/brubeck/

Cc @heathseals with whom I was talking about this

shermdog:
Superseded by #155

PR #151, opened 2016-05-20
(QENG-3919) spike for implementation of all-or-nothing checkout

rick:
:warning: do not merge :warning:

Just getting this into the system so we can look at it and discuss whether this would work, etc.

cc @joshcooper @heathseals

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -348,7 +348,7 @@ module Vmpooler
         vms = []
 
         jdata.each do |template, count|
-          val.to_i.times do |_i|
+          count.to_i.times do |_i|
             vm = fetch_single_vm(template)
             if !vm
               failed = true
@@ -366,7 +366,7 @@ module Vmpooler
         else
           vms.each do |(template, vm)|
             account_for_starting_vm(template, vm)
-            update_result_hosts(results, template, vm)
+            update_result_hosts(result, template, vm)
           end
 
           result['ok'] = true
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -86,6 +86,16 @@ module Vmpooler
       result
     end
 
+    def update_result_hosts(result, template, vm)
+      result[template] ||= {}
+      if result[template]['hostname']
+        result[template]['hostname'] = Array(result[template]['hostname'])
+        result[template]['hostname'].push(vm)
+      else
+        result[template]['hostname'] = vm
+      end
+    end
+
     get "#{api_prefix}/status/?" do
       content_type :json
 
@@ -377,16 +387,6 @@ module Vmpooler
       end
 
       JSON.pretty_generate(result)
-    end
-
-    def update_result_hosts(result, template, vm)
-      result[template] ||= {}
-      if result[template]['hostname']
-        result[template]['hostname'] = Array(result[template]['hostname'])
-        result[template]['hostname'].push(vm)
-      else
-        result[template]['hostname'] = vm
-      end
     end
 
     post "#{api_prefix}/vm/:template/?" do
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -45,35 +45,39 @@ module Vmpooler
       newhash
     end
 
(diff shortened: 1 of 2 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -335,40 +339,56 @@ def checkout_vm(template, result)
...
-      end
+        failed = false
+        vms = []
 
-      if (available == 1)
-        result['ok'] = true
-
-        jdata.each do |key, val|
+        jdata.each do |template, count|
           val.to_i.times do |_i|
```
rick: s/val/count/

spec/vmpooler/api/v1_spec.rb
```diff
@@ -265,6 +265,32 @@ def app()
         expect_json(ok = true, http = 200)
       end
 
+      it 'fails when not all requested vms can be allocated' do
+        allow(redis).to receive(:spop).with('vmpooler__ready__pool1').and_return 'abcdefghijklmnop'
```
rick: Don't really like these very mockist tests, but this is the shortest path from the current suite.

If we have a redis available, seems like we should set up state, perform operations, validate state, and then teardown state.

spec/vmpooler/api/v1_spec.rb
```diff
@@ -265,6 +265,32 @@ def app()
...
+        expected = { ok: false }
+
+        expect(last_response.body).to eq(JSON.pretty_generate(expected))
+        expect_json(ok = false, http = 200) # which HTTP status code?
+      end
+
+      it 'returns any checked out vms when not all requested vms can be allocated' do
+        allow(redis).to receive(:spop).with('vmpooler__ready__pool1').and_return 'abcdefghijklmnop'
+        allow(redis).to receive(:spop).with('vmpooler__ready__pool2').and_return nil
+        expect(redis).to receive(:spush).with("vmpooler__ready__pool1", "abcdefghijklmnop")
```
joshcooper: Do you want to test for the case where it needs to return multiple instances of the same template, and multiple instances from different templates?

joshcooper:
Just have the one question about testing N boundary cases. Otherwise looks good to me!

spec/vmpooler/api/v1_spec.rb
rick: cc #152

spec/vmpooler/api/v1_spec.rb
rick: Very much :+1: on that. I'm headed out for the day, but feel free to add those if you want.

rick:
Cool, I'll add those tests soon. We'll want to get this onto the ci.next vmpooler instance and test drive it. This will eventually need to go out to production, wondering if we have any techniques for minimizing risk, in case we missed something.

If this works as desired, what will be the impact on clients using it? We are basically going to not be returning partial sets of requested vms -- we'll either return the full set or no nodes. Seems like that shouldn't be a problem, as callers would be retrying requests in the old behavior.

cc @puppetlabs/cinext

lib/vmpooler/api/v1.rb
```diff
@@ -45,35 +45,39 @@ def alias_deref(hash)
       newhash
     end
 
-    def checkout_vm(template, result)
-      vm = backend.spop('vmpooler__ready__' + template)
+    def fetch_single_vm(template)
+      backend.spop('vmpooler__ready__' + template)
```
waynr: Does this cause messages to be emitted to the log file? If so, I can imagine the log getting very noisy during periods of high load when there could be large numbers of failures for a given vm collection request.

rick: If it does, it shouldn't happen any more often than before -- we effectively factored this out and changed the order of operations, but this isn't net new calls to spop.

joshcooper: It looks like vmpooler explicitly emits log messages:

```ruby
        $logger.log('s', "[>] [#{pool}] '#{vm}' moved to 'ready' queue")
```

So I'm thinking there shouldn't be any added verbosity. Unless our redis library also logs somewhere?

rick:
Closing in favor of #153, which is targeted to the `ci.next` branch.

PR #152, opened 2016-05-20
Improved tests for vmpooler

rick:
The tests for the vmpooler mock and stub their way to glory. The actual code is really heavy on the "use crazy redis strings everywhere" (which is mirrored in the tests). Taking a stab at making these tests be non-mockist, using actual state on the back-end, and defining the semantics of the operations we're performing.

My hope is that we get a less fragile + more readable suite in the process, and also can begin to refactor the code base, to extract and put semantic names around bits of redis-munging code scattered around (also, move the redis code out of the API into the lower `lib/` service layer(s)).

**Open issues / TODO:**

 - [x] continue de-mockistifying specs
 - [x] this probably needs to clean up test state on teardown better (or at all)
 - [x] this probably wants to use a separate redis database, so as not to clobber local redis databases that might be around. Redis uses numbered databases (what could go wrong?) still, right? Wonder if we have good examples around this we can get inspired by.

/cc @stahnma @heathseals @joshcooper


rick:
Could use some :eyes: on this now. 

I've upgraded the API tests as well as the dashboard tests to use live redis and no longer rely on mocking and stubbing.  This fixed a few tests in the process, and ends up testing the actual underlying state of the world in a few cases, where before we were mostly just checking that an HTTP status code with valid JSON were being returned.

There are a couple of files I didn't touch (in part because they seemed weird and confusing, and in part to ship something of benefit, even if not everything is covered):

 - [spec/vmpooler/pool_manager_spec.rb](https://github.com/puppetlabs/vmpooler/blob/46815f26acbad977ac89e8ed4ee5de5ec0e58557/spec/vmpooler/pool_manager_spec.rb)
 - [spec/vmpooler/api/helpers_spec.rb](https://github.com/puppetlabs/vmpooler/blob/46815f26acbad977ac89e8ed4ee5de5ec0e58557/spec/vmpooler/api/helpers_spec.rb)

The work from https://github.com/puppetlabs/vmpooler/pull/153 (which was merged to the ci.next branch) would want to merge this up and have the new tests it introduced updated once this lands. Shouldn't be difficult.


A couple of other cc's of potentially interested parties: @shermdog @heathseals

rick:
I'm going to merge my own PR unless I get a :-1:.

PR #153, opened 2016-05-23
[QENG-3919] Make vmpooler checkouts be all or nothing

rick:
Prior to this, vmpooler could partially fill requests for multiple VMs. Here we move to an algorithm that returns the entire set of requested VMs or no VMs.

/cc @puppetlabs/cinext

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gem 'rake', '>= 10.4'
 gem 'rbvmomi', '>= 1.8'
 gem 'redis', '>= 3.2'
 gem 'sinatra', '>= 1.4'
+gem 'net-ldap', '>= 0.11'
 
 # Test deps
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'rake', '>= 10.4'
 gem 'rbvmomi', '>= 1.8'
 gem 'redis', '>= 3.2'
 gem 'sinatra', '>= 1.4'
-gem 'net-ldap', '>= 0.11'
+gem 'net-ldap', '= 0.11'
 
 # Test deps
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'rake', '>= 10.4'
 gem 'rbvmomi', '>= 1.8'
 gem 'redis', '>= 3.2'
 gem 'sinatra', '>= 1.4'
-gem 'net-ldap', '= 0.11'
+gem 'net-ldap', '<= 0.12.1' # keep compatibility w/ jruby & mri-1.9.3
 
 # Test deps
 group :test do
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -72,18 +72,6 @@ module Vmpooler
       end
     end
 
-    def checkout_vm(template, result)
-      if vm = fetch_single_vm(template)
-        account_for_starting_vm(template, vm)
-        update_result_hosts(result, template, vm)
-      else
-        status 503
-        result['ok'] = false
-      end
-
-      result
-    end
-
     def update_result_hosts(result, template, vm)
       result[template] ||= {}
       if result[template]['hostname']
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -349,39 +349,11 @@ module Vmpooler
 
     post "#{api_prefix}/vm/?" do
       jdata = alias_deref(JSON.parse(request.body.read))
-
       content_type :json
       result = { 'ok' => false }
 
       if jdata and !jdata.empty?
-        failed = false
-        vms = []
-
-        jdata.each do |template, count|
-          count.to_i.times do |_i|
-            vm = fetch_single_vm(template)
-            if !vm
-              failed = true
-              break
-            else
-              vms << [ template, vm ]
-            end
-          end
-        end
-
-        if failed
-          vms.each do |(template, vm)|
-            return_single_vm(template, vm)
-          end
-        else
-          vms.each do |(template, vm)|
-            account_for_starting_vm(template, vm)
-            update_result_hosts(result, template, vm)
-          end
-
-          result['ok'] = true
-          result['domain'] = config['domain'] if config['domain']
-        end
+        result = atomically_allocate_vms(jdata)
       else
         status 404
       end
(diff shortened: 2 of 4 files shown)


shermdog:
Do we consider this an API v2 type thing?

waynr:
@shermdog my understanding is that this is addressing a bug in the v1 API so I don't think so.

briancain:
Does this deprecate the use of fetching vms through query params like what was mentioned in the ticket, or no? I don't see any spec tests around that, just fetching vms through a request body.

rick:
Good eye. I think we should probably bolster tests around that query params
use case.

rick:
Note to self: also, double-check the docs to make sure they're clear about the intended behavior.

rick:
Would actually be good to get another look at this, as I've added tests for and fixed the second code path, refactored some things and tweaked the docs.

lib/vmpooler/api/v1.rb
```diff
@@ -45,38 +45,88 @@ def alias_deref(hash)
...
+      vm = fetch_single_vm(template)
 
       unless vm.nil?
-        backend.sadd('vmpooler__running__' + template, vm)
-        backend.hset('vmpooler__active__' + template, vm, Time.now)
-        backend.hset('vmpooler__vm__' + vm, 'checkout', Time.now)
+        account_for_starting_vm(template, vm)
+        update_result_hosts(result, template, vm)
+      else
+        status 503
```
waynr: In the specs I see a comment that says "which HTTP status code?" on tests that seem (by the name) to be expecting failure and I also see `http = 200` in the `expect_json` call. Should that be `http = 503`?

lib/vmpooler/api/v1.rb
```diff
@@ -45,38 +45,88 @@ def alias_deref(hash)
...
+      else
+        result[template]['hostname'] = vm
+      end
+    end
 
-          backend.hset('vmpooler__vm__' + vm, 'token:token', request.env['HTTP_X_AUTH_TOKEN'])
-          backend.hset('vmpooler__vm__' + vm, 'token:user',
-            backend.hget('vmpooler__token__' + request.env['HTTP_X_AUTH_TOKEN'], 'user')
-          )
+    def atomically_allocate_vms(payload)
```
waynr: In the case of single-vm requests we "checkout vm" but in the case of multiple-vm requests we "atomically allocate vms"...should we try to align the semantics between these two methods to make it clear that they serve similar purposes? (the former being an operation that returns a scalar whereas the latter returns a collection) At the very least it seems like we should be able to say that the scalar operation is trivially atomic.

spec/vmpooler/api/v1_spec.rb
```diff
@@ -417,6 +531,120 @@ def app()
...
+          }
+        }
+
+        expect(last_response.body).to eq(JSON.pretty_generate(expected))
+
+        expect_json(ok = true, http = 200)
+      end
+
+      it 'returns multiple VMs even when multiple instances from multiple pools are requested' do
+        post "#{prefix}/vm/pool1+pool1+pool2+pool2+pool2", ''
```
waynr: I could be missing something, but it seems like the only difference between this new block of test and the previous new block of tests is the URL. Does RSpec allow for a test like this to be parameterized so you can tell the test case to run with each valid URL that is expected to produce the same results?

rick: It's possible to do something along those lines, but it would basically be parametrizing pairs of tests across a 6 test set, which tends to look more confusing than just having the duplication in the tests.

lib/vmpooler/api/v1.rb
rick: I added those comments and will clean them up (thanks for the reminder) ... It looks like vmpooler has always returned a 200 status, even when it cannot allocate vm's, using `return['ok']` as the indicator of whether the allocation was ok.  The 503 status is for cases where no templates were specified.

I don't 100% agree with the way it's using status + `result['ok']`, but I'm going to leave it alone, otherwise this would be a breaking change for folks.

rick: Dropped the comments in question after another round of verifying behavior.

lib/vmpooler/api/v1.rb
rick: Turns out, `checkout_vm` is no longer even used any more. :fire: :skull:

rick:
Going to go over the return status stuff one more time. I think I dropped the '503' on the floor in this change and want to make sure I get these right.

rick:
Trivia fun facts:

 - To test this stuff locally, you probably want to grab a database dump from like vmpooler-dev and install it in your local redis
 - You're going to get `{ "ok" => "false" }` results on everything unless you can auth. That means making a valid token on any test instances.

Testing results on vmpooler-dev:

```
$ curl -X POST -u rick.bradley --url vmpooler-dev.delivery.puppetlabs.net/api/v1/token -d ''
Thu May 26 15:53:45 chenrezig ~/pl/vmpooler
 (qeng-3919-all-or-nothing-checkout *%) % ▸ curl --header "X-AUTH-TOKEN: xxxx" -X POST -d '' --url vmpooler-dev.delivery.puppetlabs.net/api/v1/vm/centos-6-x86_64+centos-6-x86_64+centos-6-x86_64+centos-6-x86_64+centos-6-x86_64+centos-6-x86_64
{
  "ok": false
}Thu May 26 15:53:56 chenrezig ~/pl/vmpooler
 (qeng-3919-all-or-nothing-checkout *%) % ▸ curl --header "X-AUTH-TOKEN: xxxx" -X POST -d '' --url vmpooler-dev.delivery.puppetlabs.net/api/v1/vm/centos-6-x86_64+centos-6-x86_64+centos-6-x86_64+centos-6-x86_64
{
  "ok": true,
  "centos-6-x86_64": {
    "hostname": [
      "n21nif8dtxkbt8o",
      "b0v7kcczu4iuxwd",
      "waobk9n3d5j5ljr",
      "pcxn7yx8i6awz86"
    ]
  },
}Thu May 26 15:54:45 chenrezig ~/pl/vmpooler
 (qeng-3919-all-or-nothing-checkout *%) % ▸ curl --header "X-AUTH-TOKEN: xxxx" --url vmpooler-dev.delivery.puppetlabs.net/api/v1/vm/
[
  "centos-6-x86_64",
  "debian-7-i386",
  "debian-7-x86_64",
  "osx-1010-x86_64",
  "sles-10-x86_64",
  "solaris-10-x86_64",
  "win-2008r2-x86_64"
]Thu May 26 15:55:09 chenrezig ~/pl/vmpooler
 (qeng-3919-all-or-nothing-checkout *%) % ▸ curl --header "X-AUTH-TOKEN: xxxx" -X POST -d '' --url vmpooler-dev.delivery.puppetlabs.net/api/v1/vm/centos-6-x86_64+centos-6-x86_64+centos-6-x86_64+centos-6-x86_64+osx-1010-x86_64+osx-1010-x86_64
{
  "ok": false
}Thu May 26 15:55:40 chenrezig ~/pl/vmpooler
 (qeng-3919-all-or-nothing-checkout *%) % ▸ curl --header "X-AUTH-TOKEN: xxxx" -X POST -d '' --url vmpooler-dev.delivery.puppetlabs.net/api/v1/vm/debian-7-i386+debian-7-i386+debian-7-i386+osx-1010-x86_64+osx-1010-x86_64
{
  "ok": true,
  "debian-7-i386": {
    "hostname": [
      "spa053w9pu64jx8",
      "wpnlypog5wkq9gg",
      "fayw2046exh8icn"
    ]
  },
  "osx-1010-x86_64": {
    "hostname": [
      "uus0eaqy8we41qr",
      "pbjbfqawk0vktko"
    ]
  },
}Thu May 26 15:56:21 chenrezig ~/pl/vmpooler
 (qeng-3919-all-or-nothing-checkout *%) % ▸ curl --header "X-AUTH-TOKEN: xxxx" -X POST -d '' --url vmpooler-dev.delivery.puppetlabs.net/api/v1/vm/debian-7-i386+debian-7-i386+debian-7-i386+osx-1010-x86_64+osx-1010-x86_64
{
  "ok": false
}
Thu May 26 15:56:24 chenrezig ~/pl/vmpooler

```

waynr:
@rick it seems like those development setup gotchas should be documented somewhere, should we include that in this PR or file a ticket to write up a more thorough guide to developing on vmpooler?

rick:
@waynr yeah, no doubt.  There's a part of it (maybe specific examples) that are pretty specific to our internal network ("here, do this to grab a db dump and throw it in your local db, and then make a token and do this"), so perhaps for some of our internal wiki docs we get really specific, and for vmpooler in general we document the more general process.

rick:
As an aside, after testing this in dev I think this is good to :ship:, but one last :eyes: wouldn't hurt.

rick:
(this is also a great candidate for github's new squash merge) :tada:

Gemfile
```diff
@@ -6,6 +6,7 @@ gem 'rake', '>= 10.4'
 gem 'rbvmomi', '>= 1.8'
 gem 'redis', '>= 3.2'
 gem 'sinatra', '>= 1.4'
+gem 'net-ldap', '= 0.11'
```
rick: cc @shermdog (hard-wired to make travis happy on jruby, etc.)

waynr:
Looks good to me! :+1:

Gemfile
shermdog: Let's not hardwire as we do have some customers using this tool.

lib/vmpooler/api/v1.rb
```diff
@@ -45,38 +45,75 @@ def alias_deref(hash)
       newhash
     end
 
-    def checkout_vm(template, result)
-      vm = backend.spop('vmpooler__ready__' + template)
+    def fetch_single_vm(template)
+      backend.spop('vmpooler__ready__' + template)
+    end
+
+    def return_single_vm(template, vm)
```
shermdog: This confused me at first - could we rename it to something like re_ready_single_vm or something along those lines?

rick: Sure thing!

PR #154, opened 2016-06-02
Added IP lookup functionality for /vm/hostname

frozenfoxx:
I've discovered a need to see the IP of a vm from vmpooler when querying the status of a vm.  This code utilizes the same TCPSocket functionality found elsewhere in the codebase to retrieve the IP from vmpooler's perspective.  This information is localized to returning information about a vm since this is where all the other in-depth information about a vm is contained and only adds to the data, does not modify existing data.  This should means that scripts and workflows currently against vmpooler shouldn't need modification to continue functioning.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -459,6 +459,15 @@ module Vmpooler
           result[params[:hostname]]['disk'] = rdata['disk'].split(':')
         end
 
+        # Look up IP address of the hostname
+        begin
+          ipAddress = TCPSocket.gethostbyname(params[:hostname])[3]
+        rescue
+          ipAddress = ""
+        end
+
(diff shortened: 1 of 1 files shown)


stahnma:
👍 @rick or @shermdog ?

rick:
If someone has validated this actually works, I'm :+1: on it.

frozenfoxx:
If it helps, I've been running this internally for a few weeks without any issue, but I'd really love to see someone else run it to make sure it's okay.

shermdog:
I gave this a test and I like it!

```
{
  "ok": true,
  "mai8juy44ysam2l": {
    "template": "debian-7-x86_64",
    "lifetime": 1,
    "state": "pending",
    "ip": "10.32.121.145",
  }
}
```

PR #155, opened 2016-06-08
(RE-7014) add statsd support

shermdog:
Add the tracking of running, ready, and clone time for VM
Add the tracking of of successful, failed, invalid, and empty pool vm gets.

It is possible we may want to tweak this, but have validated with spec tests and pcaps.

```
vmpooler-tmp-dev.ready.debian-7-x86_64:1|c
vmpooler-tmp-dev.running.debian-7-x86_64:1|c

vmpooler-tmp-dev.checkout.invalid:1|c
vmpooler-tmp-dev.checkout.success.debian-7-x86_64:1|c
vmpooler-tmp-dev.checkout.empty:1|c

vmpooler-tmp-dev.running.debian-7-x86_64:1|c

vmpooler-tmp-dev.clone.debian-7-x86_64:12.10|ms

vmpooler-tmp-dev.ready.debian-7-x86_64:1|c
```

### DIFF
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -7,6 +7,7 @@ module Vmpooler
   require 'rbvmomi'
   require 'redis'
   require 'sinatra/base'
+  require "statsd-ruby"
   require 'time'
   require 'timeout'
   require 'yaml'
@@ -52,6 +53,13 @@ module Vmpooler
       parsed_config[:graphite]['prefix'] ||= 'vmpooler'
     end
 
+    # statsd is an addition and my not be present in YAML configuration
+    if parsed_config[:statsd]
+      if parsed_config[:statsd]['server']
+        parsed_config[:statsd]['prefix'] ||= 'vmpooler'
+      end
+    end
+
     if parsed_config[:tagfilter]
       parsed_config[:tagfilter].keys.each do |tag|
         parsed_config[:tagfilter][tag] = Regexp.new(parsed_config[:tagfilter][tag])
@@ -76,6 +84,14 @@ module Vmpooler
       nil
     else
       Vmpooler::Graphite.new server
+    end
+  end
+
+  def self.new_statsd(server, port)
+    if server.nil? or server.empty? or server.length == 0
+      nil
+    else
+      Statsd.new server, port
     end
   end
 
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -88,7 +88,7 @@ module Vmpooler
   end
 
   def self.new_statsd(server, port)
-    if server.nil? or server.empty? or server.length == 0
+    if server.nil? || server.empty?
       nil
     else
       Statsd.new server, port
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -12,6 +12,16 @@ module Vmpooler
       Vmpooler::API.settings.redis
     end
 
+    def statsd
+      Vmpooler::API.settings.statsd
+    end
+
+    def statsd_prefix
+      if Vmpooler::API.settings.statsd
+        Vmpooler::API.settings.config[:statsd]['prefix']? Vmpooler::API.settings.config[:statsd]['prefix'] : 'vmpooler'
+      end
+    end
+
     def config
       Vmpooler::API.settings.config[:config]
     end
@@ -32,13 +42,16 @@ module Vmpooler
       newhash = {}
 
       hash.each do |key, val|
+        if Vmpooler::API.settings.config[:alias][key]
+          key = Vmpooler::API.settings.config[:alias][key]
+        end
+
         if backend.exists('vmpooler__ready__' + key)
           newhash[key] = val
+        elsif backend.exists('vmpooler__empty__' + key)
+          newhash['empty'] = (newhash['empty'] || 0) + val.to_i
         else
-          if Vmpooler::API.settings.config[:alias][key]
-            newkey = Vmpooler::API.settings.config[:alias][key]
-            newhash[newkey] = val
-          end
+          newhash['invalid'] = (newhash['invalid'] || 0) + val.to_i
         end
       end
 
@@ -94,8 +107,10 @@ module Vmpooler
           vm = fetch_single_vm(template)
           if !vm
             failed = true
+            statsd.increment(statsd_prefix + '.checkout.fail.' + template, 1)
             break
           else
+            statsd.increment(statsd_prefix + '.checkout.success.' + template, 1)
             vms << [ template, vm ]
           end
         end
@@ -375,8 +390,16 @@ module Vmpooler
       content_type :json
       result = { 'ok' => false }
 
-      if jdata and !jdata.empty?
-        result = atomically_allocate_vms(jdata)
+      if jdata
+        empty = jdata.delete('empty')
+        invalid = jdata.delete('invalid')
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if !invalid.nil?
+        if !jdata.empty?
+          result = atomically_allocate_vms(jdata)
+        else
+          status 404
+        end
       else
         status 404
       end
@@ -400,8 +423,16 @@ module Vmpooler
       content_type :json
       result = { 'ok' => false }
 
-      if payload and !payload.empty?
-        result = atomically_allocate_vms(payload)
+      if payload
+        empty = payload.delete('empty')
+        invalid = payload.delete('invalid')
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if !invalid.nil?
+        if !payload.empty?
+          result = atomically_allocate_vms(payload)
+        else
+          status 404
+        end
       else
         status 404
       end
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -18,7 +18,7 @@ module Vmpooler
 
     def statsd_prefix
       if Vmpooler::API.settings.statsd
-        Vmpooler::API.settings.config[:statsd]['prefix']? Vmpooler::API.settings.config[:statsd]['prefix'] : 'vmpooler'
+        Vmpooler::API.settings.config[:statsd]['prefix'] ? Vmpooler::API.settings.config[:statsd]['prefix'] : 'vmpooler'
       end
     end
 
@@ -393,9 +393,9 @@ module Vmpooler
       if jdata
         empty = jdata.delete('empty')
         invalid = jdata.delete('invalid')
-        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
-        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if !invalid.nil?
-        if !jdata.empty?
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if empty
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if invalid
+        unless jdata.empty?
           result = atomically_allocate_vms(jdata)
         else
           status 404
@@ -426,9 +426,9 @@ module Vmpooler
       if payload
         empty = payload.delete('empty')
         invalid = payload.delete('invalid')
-        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
-        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if !invalid.nil?
-        if !payload.empty?
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if empty
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if invalid
+        unless payload.empty?
           result = atomically_allocate_vms(payload)
         else
           status 404
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -570,8 +570,8 @@ module Vmpooler
 
       begin
         if $statsd
-          $statsd.increment($config[:statsd]['prefix'] + '.ready.' + pool['name'], $redis.scard('vmpooler__ready__' + pool['name']))
-          $statsd.increment($config[:statsd]['prefix'] + '.running.' + pool['name'], $redis.scard('vmpooler__running__' + pool['name']))
+          $statsd.gauge($config[:statsd]['prefix'] + '.ready.' + pool['name'], $redis.scard('vmpooler__ready__' + pool['name']))
+          $statsd.gauge($config[:statsd]['prefix'] + '.running.' + pool['name'], $redis.scard('vmpooler__running__' + pool['name']))
         elsif $graphite
           $graphite.log($config[:graphite]['prefix'] + '.ready.' + pool['name'], $redis.scard('vmpooler__ready__' + pool['name']))
           $graphite.log($config[:graphite]['prefix'] + '.running.' + pool['name'], $redis.scard('vmpooler__running__' + pool['name']))
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -1,12 +1,17 @@
 module Vmpooler
   class PoolManager
-    def initialize(config, logger, redis, graphite=nil)
+    def initialize(config, logger, redis, graphite=nil, statsd=nil)
       $config = config
 
       # Load logger library
       $logger = logger
 
-      unless graphite.nil?
+      # statsd and graphite are mutex in the context of vmpooler
+      unless statsd.nil?
+        $statsd = statsd
+      end
+
+      unless graphite.nil? || !statsd.nil?
         $graphite = graphite
       end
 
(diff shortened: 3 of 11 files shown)


shermdog:
@rick please give this a :eyes:

lib/vmpooler/api/v1.rb
```diff
@@ -375,8 +390,16 @@ def atomically_allocate_vms(payload)
       content_type :json
       result = { 'ok' => false }
 
-      if jdata and !jdata.empty?
-        result = atomically_allocate_vms(jdata)
+      if jdata
+        empty = jdata.delete('empty')
+        invalid = jdata.delete('invalid')
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
```
stahnma: Why would you not use unless here instead of if not? (Is that rubocop or something)

lib/vmpooler/pool_manager.rb
```diff
@@ -1,12 +1,17 @@
 module Vmpooler
   class PoolManager
-    def initialize(config, logger, redis, graphite=nil)
+    def initialize(config, logger, redis, graphite=nil, statsd=nil)
```
stahnma: How do two optional parameters work when they are not keyword named?

lib/vmpooler/pool_manager.rb
```diff
@@ -1,12 +1,17 @@
...
       $config = config
 
       # Load logger library
       $logger = logger
 
-      unless graphite.nil?
+      # statsd and graphite are mutex in the context of vmpooler
+      unless statsd.nil?
```
stahnma: Why not one-line this? 

`$statsd = statsd unelss statsd.nil?`

lib/vmpooler/statsd.rb
```diff
@@ -0,0 +1,12 @@
+require 'rubygems' unless defined?(Gem)
+
+module Vmpooler
+  class Statsd
+    def initialize(
+      s = 'statsd',
+      port = 8125
```
stahnma: Should that be configurable somewhere?

lib/vmpooler/api/v1.rb
shermdog: Because I just follow in whatever terrible footsteps came before me :)

Learning ruby by fixing up other non-idiomatic code - what could go wrong?  :trollface:

lib/vmpooler/statsd.rb
stahnma: Nevermind, it is. Just odd that the initialize isn't one line for me :)

stahnma:
Typo in commit message.

lib/vmpooler/statsd.rb
mckern: Yeah... no reason this can't be one line and please don't use single-letter variable names. Can we add some yardoc to this so we know what these values are supposed to be? Right now I don't know if `Statsd#new` is actually expecting a string or if it expects a more complex object and this is just a placeholder.

lib/vmpooler/statsd.rb
```diff
@@ -0,0 +1,12 @@
...
+    )
+      @server = Statsd.new s, port
```
mckern: Consider wrapping the arguments to `Statsd#new` in parenthesis if you're using the return value in an assignment like this. It's functionally identical but it'll probably help avoid having to take a moment to parse the spaces here if a person has to do something here in the future.

lib/vmpooler.rb
```diff
@@ -79,6 +87,14 @@ def self.new_graphite(server)
     end
   end
 
+  def self.new_statsd(server, port)
+    if server.nil? or server.empty? or server.length == 0
```
mckern: Prefer `||` to `or`; higher binding precedence when the `if` statement is resolved. I don't think there's a need for the `server.length == 0` either -- it's identical to `server.empty?`:

```ruby
>> server = nil
=> nil
>> server.nil?
=> true
>> server.empty?
NoMethodError: undefined method `empty?' for nil:NilClass [(ripl):3:in `<main>']
>> server.length == 0
NoMethodError: undefined method `length' for nil:NilClass [(ripl):4:in `<main>']
>> server = String.new
=> ""
>> server.nil?
=> false
>> server.empty?
=> true
>> server.length == 0
=> true
```
But the real question is "When would this possibly be `nil` or an empty string?" I recommend skimming Jesse Ezell's "[IsNullOrEmpty](http://weblogs.asp.net/jezell/5280)" for more detail about why checking for `nil`or `#empty?` on an argument could be considered a code smell.

lib/vmpooler/api/v1.rb
```diff
@@ -12,6 +12,16 @@ def backend
       Vmpooler::API.settings.redis
     end
 
+    def statsd
+      Vmpooler::API.settings.statsd
+    end
+
+    def statsd_prefix
+      if Vmpooler::API.settings.statsd
+        Vmpooler::API.settings.config[:statsd]['prefix']? Vmpooler::API.settings.config[:statsd]['prefix'] : 'vmpooler'
```
mckern: Space between `['prefix']` and `?`

lib/vmpooler/api/v1.rb
mckern: `if !` should be replaced with `unless` but you're also calling `#nil?` on those values. In Ruby, all values are truthy except `nil` and `false` so you could simplify 396 and 397 some as:

```ruby
statsd.increment(statsd_prefix + '.checkout.empty', empty) if empty
statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if invalid
```

(I don't know what happens if you try to increment a statsd counter with a nil value, or else I'd say "just lost the `if` entirely")

lib/vmpooler/api/v1.rb
```diff
@@ -375,8 +390,16 @@ def atomically_allocate_vms(payload)
...
       result = { 'ok' => false }
 
-      if jdata and !jdata.empty?
-        result = atomically_allocate_vms(jdata)
+      if jdata
+        empty = jdata.delete('empty')
+        invalid = jdata.delete('invalid')
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if !invalid.nil?
+        if !jdata.empty?
```
mckern:
```ruby
unless jdata.empty?
```

lib/vmpooler/api/v1.rb
```diff
@@ -375,8 +390,16 @@ def atomically_allocate_vms(payload)
...
+          result = atomically_allocate_vms(jdata)
+        else
```
mckern: If you lose this `else; status 404`, does the next `else` (on line 404) trigger and return 404 instead?

lib/vmpooler/api/v1.rb
```diff
@@ -400,8 +423,16 @@ def extract_templates_from_query_params(params)
...
+      if payload
+        empty = payload.delete('empty')
+        invalid = payload.delete('invalid')
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if !empty.nil?
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if !invalid.nil?
+        if !payload.empty?
+          result = atomically_allocate_vms(payload)
+        else
+          status 404
+        end
```
mckern: See like, the last 3 or 4 comments re: simplifying some of lines 426 through 435 (`unless`, `#nil?`, truthiness, etc.).

lib/vmpooler/pool_manager.rb
mckern: The first three must be passed but the last two do not. They're positional.

```ruby
>> def example(a, b, c, d = nil, e = "not nil")
|    puts a
|    puts b
|    puts c
|    puts d
|    puts e
|  end
=> :example
>> example 1, 2, 3
1
2
3

not nil
=> nil
```

mckern: (Also, please throw some spaces between the value and the name in the parameter assignments. Whitespace is free.)

lib/vmpooler/pool_manager.rb
mckern: Alternatively:

```ruby
if statsd
  $statsd = statsd
elsif graphite
  $graphite = graphite
end
```

lib/vmpooler/pool_manager.rb
```diff
@@ -565,7 +571,10 @@ def _check_pool(pool)
       total = $redis.scard('vmpooler__pending__' + pool['name']) + ready
 
       begin
-        if defined? $graphite
+        if defined? $statsd
```
mckern: You don't need to use `defined?` for any of these globals because an uninitialized global variable should still return `nil` when it's called:

```ruby
# Ruby 2.1.10
>> RUBY_VERSION
=> "2.1.10"
>> RUBY_
RUBY_COPYRIGHT     RUBY_ENGINE        RUBY_PLATFORM      RUBY_REVISION
RUBY_DESCRIPTION   RUBY_PATCHLEVEL    RUBY_RELEASE_DATE  RUBY_VERSION
>> RUBY_ENGINE
=> "ruby"
>> $amidefined
=> nil
>> $noiamnot
=> nil
>>

# Ruby 2.3.1
>> RUBY_VERSION
=> "2.3.1"
>> RUBY_ENGINE
=> "ruby"
>> $amidefined
=> nil
>> $noiamnot
=> nil
>>

# Jruby 1.7.25
>> RUBY_VERSION
=> "1.9.3"
>> RUBY_ENGINE
=> "jruby"
>> $amidefined
=> nil
>> $noiamnot
=> nil
>>
```

mckern:
I left a bunch of comments, but nothing I'd consider a hard blocker. I'd probably roll all of the Graphite and Statsd magic into a single object or method that Does The Magic whenever called but we can whiteboard that out later when we start teasing some of this apart into more classes.

lib/vmpooler.rb
shermdog: I'll fix this one up, although we do expect the value to default to nil if the configuration file is invalid. https://github.com/shermdog/vmpooler/blob/8d75865a5ce034c41d41612d0d3c1dec80f1118f/vmpooler#L14

lib/vmpooler/api/v1.rb
shermdog: I "overloaded" the jdata stuff to included the empty/invalid bits.  So this is saying, after taking those bits out see if the hash is empty or not to decide if we shall proceed.

lib/vmpooler/api/v1.rb
shermdog: Nope - that's needed.

mckern: That seems smelly. I might take another look at that.

lib/vmpooler.rb
```diff
@@ -79,6 +87,14 @@ def self.new_graphite(server)
     end
   end
 
+  def self.new_statsd(server, port)
+    if server.nil? or server.empty?
```
mckern: `||` is preferred to `or`

lib/vmpooler.rb
```diff
@@ -52,6 +53,13 @@ def self.config(filepath='vmpooler.yaml')
       parsed_config[:graphite]['prefix'] ||= 'vmpooler'
     end
 
+    # statsd is an addition and my not be present in YAML configuration
+    if parsed_config[:statsd]
+      if parsed_config[:statsd]['server']
+        parsed_config[:statsd]['prefix'] ||= 'vmpooler'
+      end
+    end
```
mckern: This is defined before the code in `lib/vmpooler/api.rb` (lines 19-23), right? Are the changes to `parsed_config[:statsd]['prefix']` reflected in `Vmpooler::API.settings.config[:statsd]['prefix']`? Is there a relationship at all there? Or are we just suffering from side effects of the intermingling of model and presentation logic?

lib/vmpooler/statsd.rb
```diff
@@ -0,0 +1,9 @@
+require 'rubygems' unless defined?(Gem)
+
+module Vmpooler
+  class Statsd
+    def initialize(server = 'statsd', port = 8125)
+      @server = (Statsd.new server, port)
```
mckern: No, no, like this:

```ruby
@server = Statsd.new(server, port)
```

lib/vmpooler.rb
shermdog: Suffering.

mckern: question-about-suffering1

mckern:
flat 550x550 075 f u4

I think I'm 👍  for this!

lib/vmpooler.rb
rick: The struggle is real, and not in a flippant internet meme "my latte doesn't have a perfect rose on top" sort of sense.

lib/vmpooler/api/v1.rb
```diff
@@ -375,8 +390,16 @@ def atomically_allocate_vms(payload)
...
       result = { 'ok' => false }
 
-      if jdata and !jdata.empty?
-        result = atomically_allocate_vms(jdata)
+      if jdata
+        empty = jdata.delete('empty')
+        invalid = jdata.delete('invalid')
+        statsd.increment(statsd_prefix + '.checkout.empty', empty) if empty
+        statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if invalid
+        unless jdata.empty?
```
rick: Maybe the name I'm proposing isn't great, but I could see some encapsulation here:

``` ruby
  update_allocation_statistics(jdata)

  # ...

  def update_allocation_statistics(vm_requests)
      empty = jdata.delete('empty')
      invalid = jdata.delete('invalid')
      statsd.increment(statsd_prefix + '.checkout.empty', empty) if empty
      statsd.increment(statsd_prefix + '.checkout.invalid', invalid) if invalid
  end
```

Not blocking, could be done in a refactoring PR.

Similarly, @mckern is right that there's a (not a blocker, maybe refactoring PR) an opportunity for wider reuse in the way we access statsd.  Vague recollections, but in $past_job we had an interface that was more like:

```
  stats.increment("checkout.empty")
  stats.measure("checkout.frobnitz", 32.5)
  stats.time("checkout.duration") do
    some_time_consuming_operation(args)
  end
```

Where `stats` is a singleton-y instance of a Stats class, could be `stats` (some method/attr_reader), `$stats`, `puppetlabs.stats`, etc.

One could either specify prefix at the per-instantiation (once for All The Things) of this original thing, in which case it's prefix-relative; or, if the model is more varied, the keys are full paths. Here we're kind of doing "full paths", with prefix merged in at every use. Similarly for graphite/statsd/brubeck destination setup.

Encapsulating all the stats setup, etc., down to a class that encapsulates all the things and just proffers that 3-way interface for recording the things we need worked well for general reuse.

lib/vmpooler/pool_manager.rb
```diff
@@ -1,12 +1,15 @@
...
-    def initialize(config, logger, redis, graphite=nil)
+    def initialize(config, logger, redis, graphite = nil, statsd = nil)
       $config = config
 
       # Load logger library
       $logger = logger
 
-      unless graphite.nil?
+      # statsd and graphite are mutex in the context of vmpooler
+      if statsd
```
rick: Similarly, as above, an opportunity for refactoring; and changing the signature here if possible to not take graphite or statsd in the constructor, but internally just:

```
  def statsd
    @statsd ||= Statsd.new
  end
```

lib/vmpooler/pool_manager.rb
```diff
@@ -258,7 +261,8 @@ def clone_vm(template, folder, datastore, target)
         $redis.decr('vmpooler__tasks__clone')
 
         begin
-          $graphite.log($config[:graphite]['prefix'] + ".clone.#{vm['template']}", finish) if defined? $graphite
+          $statsd.timing($config[:statsd]['prefix'] + ".clone.#{vm['template']}", finish) if $statsd
```
rick: And I think we want to always presume statsd is available, and unify statsd/graphite into one thing if possible. Fire and forget. If it's botched then log/warn (inside our wrapper class), but carry on and not interrupt the caller.

In contexts where it made sense we often just published some data via activesupport notifications or whatever and stats became a listener; not recommended here.

spec/vmpooler/api/v1_spec.rb
```diff
@@ -180,6 +180,7 @@ def app()
 
   describe '/vm' do
     let(:redis)  { double('redis') }
+    let(:statsd)  { double('stats') }
```
rick: If we have a class representing statsd we could actually just subclass is in testing and neuter the publication / record in the statsd object that things were to be published.  It's the same as a double or mock, in many ways, but just mentioning it (because "such mock")

spec/vmpooler/api/v1_spec.rb
```diff
@@ -190,20 +191,27 @@ def app()
...
+      allow(redis).to receive(:exists).with('vmpooler__ready__pool2').and_return '1'
       allow(redis).to receive(:hget).with('vmpooler__token__abcdefghijklmnopqrstuvwxyz012345', 'user').and_return 'jdoe'
       allow(redis).to receive(:hset).and_return '1'
       allow(redis).to receive(:sadd).and_return '1'
       allow(redis).to receive(:scard).and_return '5'
       allow(redis).to receive(:spop).with('vmpooler__ready__pool1').and_return 'abcdefghijklmnop'
       allow(redis).to receive(:spop).with('vmpooler__ready__pool2').and_return 'qrstuvwxyz012345'
+      allow(statsd).to receive(:increment).with('vmpooler.checkout.success.pool1', 1)
+      allow(statsd).to receive(:increment).with('vmpooler.checkout.success.pool2', 1)
+      allow(statsd).to receive(:increment).with('vmpooler.checkout.fail.pool2', 1)
```
rick: A testing subclass of statsd would just not publish, would get real calls, and in cases like this we wouldn't have to "allow" (which is brittle under change); when we want to assert collection, we could ask the test subclass if it got the things we expected.

rick:
Many comments related to refactoring directions. Not blocking.

:+1:

PR #158, opened 2016-07-08
(QENG-4070) Consistently return 503 if valid pool is empty

joshcooper:
This merges the most recent code in `master`, which adds some much needed non-mockist tests.

It reverts RE-7014, as that had an unattended side-effect that you *must* specify an alias in your vmpooler config file. If you don't have that in your config, vmpooler will try to call a method on nil and blow up. The changes also relied on the `alias_deref` method, which conflated "is this a valid pool" with "is the pool empty". This PR straightens that out and should make it easier to reapply the stats changes.

It also updates vmpooler to correctly differentiate between an empty pool and an invalid pool name, and consistently returns 503 in the former and 404 in the latter.


rick:
So, ruby 1.9.* has the `json` library in stdlib so it shouldn't be included in the Gemfile, as I understand it.  Which is fine and all. But not sure why this test wasn't failing on travis before.

Seems like all the tests are green on the other platforms, which is :metal:.

joshcooper:
The `json` failure on 1.9 is because there was a new `json` version published July 1 and it requires ruby 2, see https://rubygems.org/gems/json/versions/2.0.1. I don't understand why bundler is choosing to install that version since it knows what version of ruby it's running in, and the `json` gem requires ruby 2:

```
required_ruby_version: !ruby/object:Gem::Requirement
  requirements:
  - - "~>"
    - !ruby/object:Gem::Version
      version: '2.0'
```

I'll add a commit to restrict json to `~> 1.8` which is what we were using prior to July 1.

PR #159, opened 2016-07-11
[QENG-4075] Land statsd support after earlier ci.next changes

rick:
Getting statsd support landed back onto the ci.next branch after earlier bugfixes in #158 

**todo**

- [x] re-apply statsd PR (revert the revert)
- [x] simplify statsd call points using new code paths
- [x] fix broken tests
- [x] diagnose statsd `nil` problems on startup
- [x] any 💄 / cleanups / refactorings
- [x] refactor graphite code to use uniform API
- [x] refactor metrics initialization code to be uniform
- [x] rename `statsd` to `metrics` where it is used
- [x] graphite server configuration is used on the dashboard to display our graphs :warning:
- [x] test all the things:
  - [x] ensure no breakage on old config:
    - [x] multiple checkout works
    - [x] exhausting pools returns 503, with multiple checkouts
    - [x] single checkout works
    - [x] exhausting pools returns 503, with single checkouts
    - [x] invalid pools return 404
    - [x] alias works
    - [x] alias checkout returns 503 on exhausted pools   
  - [x] no configuration (graphite, graphs, statsd)
  - [x] graphite configuration only (publishing stats, graphs loading)
  - [x] graphs configuration only (no publishing, graphs loading from graphs)
  - [x] statsd configuration only (publishing stats, no graphs)
  - [x] statsd configuration + graphite configuration (publishes to statsd, gets graphs from graphite)
  - [x] statsd + graphs (publishes to statsd, gets graphs from graphs)
  - [x] graphite + graphs (publishes to graphite, gets graphs from graphs)
  - [x] statsd + graphite + graphs (publishes to statsd, gets graphs from graphs)
- [x] :eyes:

cc  @puppetlabs/vmpooler


rick:
**Note**:

I've found that to get this to run on the ci.next vmpooler instance, I have to do:

```
$ /usr/bin/jruby -S gem install statsd-ruby
```

Otherwise the `service` triggered instance and the run-it-by-hand-with `/usr/bin/jruby` instance will fail with an error message about not being able to require the `statsd` library.

I'm far from a jruby expert, but it seems like we are not using bundler with our jruby vmpooler installations. I wonder if this is just a local problem (i.e. not all jruby users would face it), but it's sort of hard-coded into our run script.

Should we be doing `jruby -S bundle install` and `jruby -S bundle exec vmpooler ...`?

rick:
This needs more manual testing. I have the vmpooler.yaml set up on the ci.next vmpooler instance and was going to walk through the testing checkboxes up there ⬆️ with different config settings; do requests, make sure nothing explodes, see if stats are recorded to reasonable places.

But I have to run 🏠. So, later.

stahnma:
@puppetlabs/vmpooler check this out!

rick:
I've tested all the various configurations for this and everything seems to work. We're getting graphs for the dashboard in all the right places, given the configs; and we're publishing to all the right endpoints, given the configs. I'm sure there's something I missed testing, but it seems to be working from what I've been able to drive.

Would definitely appreciate some :eyes: on this.

:rocket:

lib/vmpooler/api/v1.rb
```diff
@@ -372,8 +376,16 @@ def atomically_allocate_vms(payload)
...
+        if invalid.empty?
+          result = atomically_allocate_vms(payload)
+        else
+          invalid.each do |bad_template|
+            metrics.increment('checkout.invalid.' + bad_template)
+          end
+          status 404
+        end
       else
         status 404
```
joshcooper: Should we increment an `invalid` counter? Since payload is nil, do we lump them into the same bucket, e.g. `metrics.increment('checkout.invalid.<unknown>')`?

lib/vmpooler/api/v1.rb
```diff
@@ -406,8 +418,16 @@ def all_templates_valid?(payload)
...
```
joshcooper: same question about `metrics.increment`?

vmpooler.yaml.example
```diff
@@ -53,20 +53,79 @@
...
-# historical data in an external Graphite database.
+# historical data in an external Graphite database.  This is mutually exclusive
+# with statsd.
 #
 # Available configuration parameters:
 #
 #   - server
 #     The FQDN hostname of the Graphite server.
-#     (optional)
+#     (required)
```
joshcooper: Just noting that we claimed `server` was optional, but it wasn't really due to https://github.com/puppetlabs/vmpooler/pull/159/files#diff-39bf7d15d9519a285b183e3fa9505448L11 and https://github.com/puppetlabs/vmpooler/pull/159/files#diff-e64c536b990385f25acd67c743ad6574L79

rick: I have it here as "required". I guess that's a behavior change (though I wonder if anyone is really relying on the default behavior).

joshcooper: I think what you have is good. I don't think the parameter was optional in practice due to the lines above.

lib/vmpooler/api/v1.rb
rick: I think that would be useful.  I believe "<"/">" aren't valid for graphite metric names, fwiw. Probably just "unknown" would do (if someone gets confused about a pool they have named "unknown", I'm not sure what to tell them).

shermdog:
I had no idea we were getting graphs on the dashboard from graphite.  @charlesdunbar do we need to take anything into consideration for the graphs once we start exclusively using statsd?

charlesdunbar:
Considerations would be the same with statsd as with any metric in graphite.  Ensuring the naming scheme matches how often graphite is configured to expect data, any rollups/aggregations happen as often or rarely as you'd like, and if there are rollups/aggregations, the correct method is being applied to the data.

All those settings are configurable in our puppetlabs-modules repo - it may help to discuss what metrics are being sent and what questions you'd like answered from the data.

shermdog:
I am good with merging this.  I tested manually and confirmed the stats are going out as we expect and graphs are showing up.

```
vmpooler-tmp-dev.checkout.invalid.bad:1|c
vmpooler-tmp-dev.ready.debian-7-x86_64:1|g
vmpooler-tmp-dev.running.debian-7-x86_64:0|g
vmpooler-tmp-dev.checkout.success.debian-7-x86_64:1|c
vmpooler-tmp-dev.ready.debian-7-x86_64:0|g
vmpooler-tmp-dev.running.debian-7-x86_64:1|g
vmpooler-tmp-dev.clone.debian-7-x86_64:6.54|ms
vmpooler-tmp-dev.ready.debian-7-x86_64:1|g
vmpooler-tmp-dev.running.debian-7-x86_64:1|g
```

PR #160, opened 2016-07-20
(QENG-3376) Allow lifetime to be set per-pool

shermdog:
At current vm_lifetime and vm_lifetime_auth are configured globally only.  This change allows that value to be set/overridden per pool.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -66,8 +66,9 @@ module Vmpooler
           backend.hget('vmpooler__token__' + request.env['HTTP_X_AUTH_TOKEN'], 'user')
         )
 
-        if config['vm_lifetime_auth'].to_i > 0
-          backend.hset('vmpooler__vm__' + vm, 'lifetime', config['vm_lifetime_auth'].to_i)
+        pool_lifetime_auth = pools.find {|p| p["name"] == template }['vm_lifetime_auth']
+        if (pool_lifetime_auth || config['vm_lifetime_auth']).to_i > 0
+          backend.hset('vmpooler__vm__' + vm, 'lifetime', (pool_lifetime_auth || config['vm_lifetime_auth']).to_i)
         end
       end
     end
@@ -455,13 +456,14 @@ module Vmpooler
         result[params[:hostname]] = {}
 
         result[params[:hostname]]['template'] = rdata['template']
-        result[params[:hostname]]['lifetime'] = (rdata['lifetime'] || config['vm_lifetime']).to_i
+        pool_auth = pools.find {|p| p["name"] == rdata['template'] }['vm_lifetime']
+        result[params[:hostname]]['lifetime'] = (rdata['lifetime'] || pool_auth || config['vm_lifetime']).to_i
 
         if rdata['destroy']
-          result[params[:hostname]]['running'] = ((Time.parse(rdata['destroy']) - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          result[params[:hostname]]['running'] = ((Time.parse(rdata['destroy']) - Time.parse(rdata['checkout'])) / 60 / 60).round(2).to_f
           result[params[:hostname]]['state'] = 'destroyed'
         elsif rdata['checkout']
-          result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2).to_f
           result[params[:hostname]]['state'] = 'running'
         elsif rdata['check']
           result[params[:hostname]]['state'] = 'ready'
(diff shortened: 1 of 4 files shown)


stahnma:
@heathseals here's another vmpooler item you're welcome to take a look at.

stahnma:
so the tests fail because 0 != 0.0. We should probably compare the parsed json vs the string. Anyway, @mckern will pair with somebody on fixing tests. /cc @heathseals or @shermdog

mckern:
We're going to hack on this the week of 08-29 when @shermdog is in town

lib/vmpooler/api/v1.rb
```diff
@@ -66,8 +66,9 @@ def account_for_starting_vm(template, vm)
           backend.hget('vmpooler__token__' + request.env['HTTP_X_AUTH_TOKEN'], 'user')
         )
 
-        if config['vm_lifetime_auth'].to_i > 0
-          backend.hset('vmpooler__vm__' + vm, 'lifetime', config['vm_lifetime_auth'].to_i)
+        pool_lifetime_auth = pools.find {|p| p["name"] == template }['vm_lifetime_auth']
```
mckern: nitpick: `{ |p|` instead of `{|p|`?

lib/vmpooler/api/v1.rb
```diff
@@ -66,8 +66,9 @@ def account_for_starting_vm(template, vm)
...
+        if (pool_lifetime_auth || config['vm_lifetime_auth']).to_i > 0
+          backend.hset('vmpooler__vm__' + vm, 'lifetime', (pool_lifetime_auth || config['vm_lifetime_auth']).to_i)
```
mckern: You're calling `(pool_lifetime_auth || config['vm_lifetime_auth'])` enough that it might be worth moving that into an accessor or a helper method:

```ruby
# Stick this somewhere useful, maybe where `#pool_lifetime_auth` is defined?
# @return [Float, nil] the length of time a VM has been active
def auth_lifetime
  pool_lifetime_auth || config['vm_lifetime_auth']
end

if auth_lifetime.to_i > 0
  backend.hset('vmpooler__vm__' + vm, 'lifetime', auth_lifetime.to_i)
end
```

This is kind of boilerplate, but it's something to consider. I'm OK with merging this and then implementing something like what I've suggested in another PR.

lib/vmpooler/api/v1.rb
```diff
@@ -455,13 +456,14 @@ def invalid_templates(payload)
         result[params[:hostname]] = {}
 
         result[params[:hostname]]['template'] = rdata['template']
-        result[params[:hostname]]['lifetime'] = (rdata['lifetime'] || config['vm_lifetime']).to_i
+        pool_auth = pools.find {|p| p["name"] == rdata['template'] }['vm_lifetime']
```
mckern: More style nitpick: `{ |p|` instead of  `{|p|`; obviously non-blocking

lib/vmpooler/api/v1.rb
```diff
@@ -455,13 +456,14 @@ def invalid_templates(payload)
...
+        result[params[:hostname]]['lifetime'] = (rdata['lifetime'] || pool_auth || config['vm_lifetime']).to_i
 
         if rdata['destroy']
-          result[params[:hostname]]['running'] = ((Time.parse(rdata['destroy']) - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          result[params[:hostname]]['running'] = ((Time.parse(rdata['destroy']) - Time.parse(rdata['checkout'])) / 60 / 60).round(2).to_f
```
mckern: We do a lot of date math throughout this file; we should move that date math into some helper functions. Probably out of scope for this PR.

mckern:
I left some nitpicks/comments but I'm 👍 to merging this and using the comments for future maintainaince.

lib/vmpooler/api/v1.rb
rick: :+1: on @mckern's suggestion here. There's a lot of copy-pasta inside vmpooler and I think we're well served by extracting methods with intention-revealing names when we can.

lib/vmpooler/api/v1.rb
rick: :+1: as well.  Anything that's a string of math or `||`s probably could have a good name and be pulled out.

spec/helpers.rb
```diff
@@ -36,7 +36,7 @@ def create_ready_vm(template, name, token = nil)
   create_vm(name, token)
   redis.sadd("vmpooler__ready__#{template}", name)
   # REMIND: should be __vm__?
-  redis.hset("vmpooler_vm_#{name}", "template", template)
+  redis.hset("vmpooler__vm__#{name}", "template", template)
```
rick: Good catch.

rick: Outside the scope of this PR, but my hope has been that we eventually refactor enough of the core parts of the pooler into methods (or ideally classes with methods) that hide the redis-y bits, and then these spec helpers just disappear in favor of using the real domain model.

PR #167, opened 2016-10-19
Improve vmpooler scheduling logic

mattkirby:
This change implements improved scheduling of pooled VMs by updating the logic used to find the least used host in a target cluster. Without this change host selection is based on VM count, which does not give an indication of VM size or current host utilization.

Additionally, this change implements a queue for migrating VMs at request time. The least used compatible host is selected for migration, which occurs transparently to the consumer, and migration is performed if required.

Without these changes vmpooler may cause uneven load distribution across a cluster of resources since the consumption of VMs is separated from scheduling.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -478,7 +478,7 @@ module Vmpooler
 
     def migration_limit(migration_limit)
       # Returns migration_limit setting when enabled
-      return false if migration_limit == 0 or not migration_limit
+      return false if migration_limit == 0 || ! migration_limit
       migration_limit if migration_limit >= 1
     end
 
@@ -494,13 +494,15 @@ module Vmpooler
         vm_object = find_vsphere_pool_vm(pool, vm, vsphere)
         parent_host, parent_host_name = get_vm_host_info(vm_object)
         migration_limit = migration_limit $config[:config]['migration_limit']
+        migration_count = $redis.scard('vmpooler__migration')
 
-        if not migration_limit
+        if ! migration_limit
           $logger.log('s', "[ ] [#{pool}] '#{vm}' is running on #{parent_host_name}")
+          return
         else
-          migration_count = $redis.smembers('vmpooler__migration').size
           if migration_count >= migration_limit
             $logger.log('s', "[ ] [#{pool}] '#{vm}' is running on #{parent_host_name}. No migration will be evaluated since the migration_limit has been reached")
+            return
           else
             $redis.sadd('vmpooler__migration', vm)
             host, host_name = vsphere.find_least_used_compatible_host(vm_object)
@@ -585,7 +587,8 @@ module Vmpooler
       $redis.smembers("vmpooler__running__#{pool['name']}").each do |vm|
         if inventory[vm]
           begin
-            check_running_vm(vm, pool['name'], $redis.hget('vmpooler__vm__' + vm, 'lifetime') || $config[:config]['vm_lifetime'] || 12, vsphere)
+            vm_lifetime = $redis.hget('vmpooler__vm__' + vm, 'lifetime') || $config[:config]['vm_lifetime'] || 12
+            check_running_vm(vm, pool['name'], vm_lifetime, vsphere)
           rescue
           end
         end
@@ -605,7 +608,8 @@ module Vmpooler
       $redis.smembers("vmpooler__pending__#{pool['name']}").each do |vm|
         if inventory[vm]
           begin
-            check_pending_vm(vm, pool['name'], pool['timeout'] || $config[:config]['timeout'] || 15, vsphere)
+            pool_timeout = pool['timeout'] || $config[:config]['timeout'] || 15
+            check_pending_vm(vm, pool['name'], pool_timeout, vsphere)
           rescue
           end
         end
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -26,29 +26,30 @@ module Vmpooler
       end
     end
 
-    def open_socket(host, domain=nil, timeout=5, port=22)
+    def open_socket(host, domain=nil, timeout=5, port=22, &block)
       Timeout.timeout(timeout) do
-        target_host = vm
-        target_host = "#{vm}.#{domain}" if domain
-        TCPSocket.new target_host, port
+        target_host = host
+        target_host = "#{host}.#{domain}" if domain
+        sock = TCPSocket.new target_host, port
+        begin
+          yield sock if block_given?
+        ensure
+          sock.close
+        end
       end
     end
 
     def _check_pending_vm(vm, pool, timeout, vsphere)
       host = vsphere.find_vm(vm)
 
-      if host
-        begin
-          Timeout.timeout(5) do
-            TCPSocket.new vm, 22
-          end
-          move_pending_vm_to_ready(vm, pool, host)
-        rescue
-          fail_pending_vm(vm, pool, timeout)
-        end
-      else
-        fail_pending_vm(vm, pool, timeout)
+      if ! host
+        fail_pending_vm(vm, pool, timeout, false)
+        return
       end
+      open_socket vm
+      move_pending_vm_to_ready(vm, pool, host)
+    rescue
+      fail_pending_vm(vm, pool, timeout)
     end
 
     def fail_pending_vm(vm, pool, timeout, exists=true)
@@ -137,12 +138,12 @@ module Vmpooler
           end
 
           begin
-            Timeout.timeout(5) do
-              TCPSocket.new vm, 22
-            end
+            open_socket vm
           rescue
             if $redis.smove('vmpooler__ready__' + pool, 'vmpooler__completed__' + pool, vm)
               $logger.log('d', "[!] [#{pool}] '#{vm}' is unreachable, removed from 'ready' queue")
+            else
+              $logger.log('d', "[!] [#{pool}] '#{vm}' is unreachable, and failed to remove from 'ready' queue")
             end
           end
         end
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -455,6 +455,55 @@ module Vmpooler
       end
     end
 
+    def find_vsphere_pool_vm(pool, vm)
+      $vsphere[pool].find_vm(vm) || $vsphere[pool].find_vm_heavy(vm)[vm]
+    end
+
+    def migration_enabled?(migration_limit)
+      # Returns migration_limit setting when enabled
+      return false if migration_limit == 0 or not migration_limit
+      migration_limit if migration_limit >= 1
+    end
+
+    def migrate_vm(vm, pool)
+      Thread.new do
+        _migrate_vm(vm, pool)
+      end
+    end
+
+    def _migrate_vm(vm, pool)
+      $redis.srem('vmpooler__migrating__' + pool, vm)
+      vm_object = find_vsphere_pool_vm(pool, vm)
+      parent_host = vm_object.summary.runtime.host
+      parent_host_name = parent_host.name
+      migration_limit = migration_enabled? $config[:config]['migration_limit']
+
+      if not migration_limit
+        $logger.log('s', "[ ] [#{pool}] '#{vm}' is running on #{parent_host_name}")
+      else
+        migration_count = $redis.smembers('vmpooler__migration').size
+        if migration_count >= migration_limit
+          $logger.log('s', "[ ] [#{pool}] '#{vm}' is running on #{parent_host_name}. No migration will be evaluated since the migration_limit has been reached")
+        else
+          $redis.sadd('vmpooler__migration', vm)
+          host = $vsphere[pool].find_least_used_compatible_host(vm_object)
+          if host == parent_host
+            $logger.log('s', "[ ] [#{pool}] No migration required for '#{vm}' running on #{parent_host_name}")
+          else
+            start = Time.now
+            $vsphere[pool].migrate_vm_host(vm_object, host)
+            finish = '%.2f' % (Time.now - start)
+            $metrics.timing("migrate.#{vm['template']}", finish)
+            checkout_to_migration = '%.2f' % (Time.now - Time.parse($redis.hget('vmpooler__vm__' + vm, 'checkout')))
+            $redis.hset('vmpooler__vm__' + vm, 'migration_time', finish)
+            $redis.hset('vmpooler__vm__' + vm, 'checkout_to_migration', checkout_to_migration)
+            $logger.log('s', "[>] [#{pool}] '#{vm}' migrated from #{parent_host_name} to #{host.name} in #{finish} seconds")
+          end
+          $redis.srem('vmpooler__migration', vm)
+        end
+      end
+    end
+
     def check_pool(pool)
       $logger.log('d', "[*] [#{pool['name']}] starting worker thread")
 
@@ -480,7 +529,8 @@ module Vmpooler
             (! $redis.sismember('vmpooler__ready__' + pool['name'], vm['name'])) &&
             (! $redis.sismember('vmpooler__pending__' + pool['name'], vm['name'])) &&
             (! $redis.sismember('vmpooler__completed__' + pool['name'], vm['name'])) &&
-            (! $redis.sismember('vmpooler__discovered__' + pool['name'], vm['name']))
+            (! $redis.sismember('vmpooler__discovered__' + pool['name'], vm['name'])) &&
+            (! $redis.sismember('vmpooler__migrating__' + pool['name'], vm['name']))
 
             $redis.sadd('vmpooler__discovered__' + pool['name'], vm['name'])
 
@@ -552,6 +602,17 @@ module Vmpooler
 
         if $redis.sismember('vmpooler__discovered__' + pool['name'], vm)
           $redis.smove('vmpooler__discovered__' + pool['name'], 'vmpooler__completed__' + pool['name'], vm)
+        end
+      end
+
+      # MIGRATIONS
+      $redis.smembers('vmpooler__migrating__' + pool['name']).each do |vm|
+        if inventory[vm]
+          begin
+            migrate_vm(vm, pool['name'])
+          rescue => err
+            $logger.log('s', "[x] [#{pool['name']}] '#{vm}' failed to migrate: #{err}")
+          end
         end
       end
 
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -49,16 +49,21 @@ module Vmpooler
       end
     end
 
-    def fail_pending_vm(vm, pool, timeout)
-      clone_stamp = $redis.hget('vmpooler__vm__' + vm, 'clone')
+    def fail_pending_vm(vm, pool, timeout, exists=true)
+      clone_stamp = $redis.hget("vmpooler__vm__#{vm}", 'clone')
+      return if ! clone_stamp
 
-      if (clone_stamp) &&
-          (((Time.now - Time.parse(clone_stamp)) / 60) > timeout)
-
-        $redis.smove('vmpooler__pending__' + pool, 'vmpooler__completed__' + pool, vm)
-
-        $logger.log('d', "[!] [#{pool}] '#{vm}' marked as 'failed' after #{timeout} minutes")
+      time_since_clone = (Time.now - Time.parse(clone_stamp)) / 60
+      if time_since_clone > timeout
+        if exists
+          $redis.smove('vmpooler__pending__' + pool, 'vmpooler__completed__' + pool, vm)
+          $logger.log('d', "[!] [#{pool}] '#{vm}' marked as 'failed' after #{timeout} minutes")
+        else
+          remove_nonexistent_vm(vm, pool)
+        end
       end
+    rescue => err
+      $logger.log('d', "Fail pending VM failed with an error: #{err}")
     end
 
     def move_pending_vm_to_ready(vm, pool, host)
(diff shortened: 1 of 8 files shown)


mattkirby:
This change is ready for review, but is blocked from moving forward until I present the changes next week. Please feel free to offer any review in the interim.

spec/vmpooler/pool_manager_spec.rb
```diff
@@ -228,6 +228,7 @@
       allow(redis).to receive(:smembers).with('vmpooler__running__pool1').and_return([])
       allow(redis).to receive(:smembers).with('vmpooler__completed__pool1').and_return([])
       allow(redis).to receive(:smembers).with('vmpooler__discovered__pool1').and_return([])
+      allow(redis).to receive(:smembers).with('vmpooler__migrating__pool1').and_return([])
```
rick: The pool manager has the only remaining hive of "mockist" specs in the pooler (I converted the remaining specs over to non-mockist specs).  The "mockist" specs don't really exercise much of anything, but I stopped short here because rewriting these specs is a pain.

It would not be a blocker for this PR to not put in some real specs for the new functionality, but I would definitely give big :sparkles: / :tada: to the effort.

If you end up wanting to go that route, there are some helpers in https://github.com/puppetlabs/vmpooler/blob/master/spec/helpers.rb that are designed to set up actual vm state in tests, which allows us to then call our methods to manipulate and then assert that things look right in redis.

For example, here's a fairly stateful non-mockist spec: https://github.com/puppetlabs/vmpooler/blob/master/spec/vmpooler/api/v1/vm_template_spec.rb#L173-L184

Anyway, don't worry about it if it's a pain (it probably is).  It's on our (my?) wishlist at some point to have our suite be more real, and to evolve the pool manager so that it has a reasonable domain model reflected in the code and those test helpers (or something similar) migrate down into the code itself.

lib/vmpooler/pool_manager.rb
```diff
@@ -455,6 +455,31 @@ def _check_snapshot_queue
...
 
+    def migrate_vm(vm, pool)
+      Thread.new do
+        _migrate_vm(vm, pool)
+      end
+    end
+
+    def _migrate_vm(vm, pool)
+      $redis.srem('vmpooler__migrating__' + pool, vm)
+      vm_object = $vsphere[pool].find_vm(vm) || $vsphere[pool].find_vm_heavy(vm)
```
rick: Again (see below), not a priority to fix here, but if it's easy, this incantation could become a method with a name (e.g., `find_vsphere_pool_vm(pool, vm)`).

lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +581,17 @@ def _check_pool(pool)
         end
       end
 
+      # MIGRATIONS
+      $redis.smembers('vmpooler__migrating__' + pool['name']).each do |vm|
+        if inventory[vm]
+          begin
+            migrate_vm(vm, pool['name'])
+          rescue
```
rick: It might be handy for logging purposes here to make this:

``` ruby
begin
  migrate_vm(vm, pool['name'])
rescue => err
  $logger.log('s', '[x] [' + pool['name'] + "] '" + vm + "' failed to migrate: #{err}")
end
```

The log line could also be written as:

``` ruby
$logger.log('s', "[x] [#{pool['name']}] '#{vm}' failed to migrate: #{err}")
```

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,31 +196,62 @@ def find_folder(foldername)
...
+          memory_utilization = (memory_usage.to_f / memory_size.to_f) * 100
+          if not cpu_utilization > limit and not memory_utilization > limit
+            utilization = cpu_utilization + memory_utilization
+            return [ utilization, host ]
+          end
+        end
+      end
+      return nil
+    end
+
```
rick: A potential refactoring of this method (untested):

``` ruby
  def get_host_utilization(host, model=nil, limit=90)
    return nil unless model
    return nil unless host_has_cpu_model? host, model
    return nil if host.runtime.inMaintenanceMode
    return nil unless host.overallStatus == 'green'

    cpu_utilization = cpu_utilization_for host
    memory_utilization = memory_utilization_for host

    return nil if cpu_utilization > limit
    return nil if memory_utilization > limit

    [ cpu_utilization + memory_utilization, host ]
  end

  def host_has_cpu_model?(host, model)
    host.hardware.cpuPkg[0].description.split().include? model
  end

  def cpu_utilization_for(host)
    cpu_usage = host.summary.quickStats.overallCpuUsage
    cpu_size = host.summary.hardware.cpuMhz * host.summary.hardware.numCpuCores
    (cpu_usage.to_f / cpu_size.to_f) * 100
  end

  def memory_utilization_for(host)
    memory_usage = host.summary.quickStats.overallMemoryUsage
    memory_size = host.summary.hardware.memorySize / 1024 / 1024
    (memory_usage.to_f / memory_size.to_f) * 100
  end
```

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,31 +196,62 @@ def find_folder(foldername)
...
+          host_usage = get_host_utilization(host)
+          target_hosts << host_usage if host_usage
         end
       end
+      return target_hosts.sort[0][1]
+    end
 
-      hosts[hosts_sort.sort_by { |_k, v| v }[0][0]]
+    def find_least_used_compatible_host(vm)
+      begin
```
rick: This block is a bit cryptic -- I think we're basically calling a thing for the side effect of throwing an error if uninitialized and then initializing in that case.  Probably could extract it to a method and give it an intention-revealing name?

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,31 +196,62 @@ def find_folder(foldername)
...
+        @connection.serviceInstance.CurrentTime
+      rescue
+        initialize
+      end
+
+      source_host = vm.summary.runtime.host
+      model = source_host.hardware.cpuPkg[0].description.split()[4]
```
rick: This bit: `host.hardware.cpuPkg[0].description.split()` turns up in multiple places. What is this thing? Can we use the name for the meaning of it, wrap it in a method, and call it?  The `[4]` bit is a bit cryptic. If we extract to a method we could document there what the structure is. Maybe even make a method with a clear name that does the `[4]` part.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,31 +196,62 @@ def find_folder(foldername)
...
+      cluster = source_host.parent
+      target_hosts = []
+      cluster.host.each do |host|
+        host_usage = get_host_utilization(host, model)
+        target_hosts << host_usage if host_usage
+      end
+      return target_hosts.sort[0][1]
```
rick: Nitpick: we don't need `return` at the end of a method in ruby, it's enough just to say `target_hosts.sort[0][1]`.

Also, I notice that we're doing similar things here as we are higher up in the diff.  Makes me think there's a more fundamental operation happening that we could extract to a method and share?

rick:
Mostly style / refactoring / readability / nitpick comments.  Nothing that I mentioned would block this PR.  Am totally game to pair on this if you want to at some point.

spec/vmpooler/pool_manager_spec.rb
mattkirby: I do want to setup some real tests for the redis usage as I think it will uncover some bugs. I would like to move this change forward separately from that, but I am definitely happy to put in some time to do this since I think it will help identify some of the problem areas of the current redis usage and increase confidence in future changes. The examples are very helpful for me since I have not written much in the way of rub

lib/vmpooler/pool_manager.rb
mattkirby: I updated this as suggested.

lib/vmpooler/pool_manager.rb
```diff
@@ -455,6 +455,31 @@ def _check_snapshot_queue
...
+    def _migrate_vm(vm, pool)
+      $redis.srem('vmpooler__migrating__' + pool, vm)
+      vm_object = $vsphere[pool].find_vm(vm) || $vsphere[pool].find_vm_heavy(vm)
+      host = $vsphere[pool].find_least_used_compatible_host(vm_object)
+      parent_host = vm_object.summary.runtime.host
+      parent_host_name = parent_host.name
+      if host == parent_host
+        $logger.log('s', '[ ] [' + pool + "] No migration required for '" + vm + "'")
+      else
+        start = Time.now
```
mattkirby: This has been added. It looks like it follows the currently used patterns, but please let me know if I've missed something or if it needs further adjustment.

lib/vmpooler/pool_manager.rb
mattkirby: I have added the context of the failure to the log message and cleaned up the logger string.

lib/vmpooler/vsphere_helper.rb
mattkirby: I ran through and put together something that looks nearly like this and still does the same thing. I think it captures what you were suggesting and helps clean it up. Thank you.

lib/vmpooler/vsphere_helper.rb
mattkirby: Yes. It's a existing pattern that I duplicated here that tries to see if a connection is alive and available for reuse. I moved it to a method and replaced usage of this pattern with a call of the method.

lib/vmpooler/vsphere_helper.rb
mattkirby: I've made an attempt to clarify this with methods. I think it should be more clear now, but am happy to increase clarity if it still requires improvement.

lib/vmpooler/pool_manager.rb
```diff
@@ -455,6 +455,31 @@ def _check_snapshot_queue
       end
     end
 
+    def migrate_vm(vm, pool)
+      Thread.new do
```
mattkirby: Yes, a new thread for each VM that needs to be migrated. The limitation ends up being placed by the number of vsphere connections, and there are shared vsphere connections per pool that end up being used (I'm not sure how many). There's a concept of limiting concurrent tasks in vmpooler, but I haven't gotten into the nuts and bolts of how that works yet. I suspect there is some room for optimization. It looks like the model of vmpooler is to always create more threads, so this is consistent with how the application is handling clone operations, and other actions of that nature. The number of operations able to be performed at any one time will always be limited by the number of vsphere connections.

Since the question of which host is the best target for my VM right now is asked for every VM, and the answer is not shared between migration jobs, then it should be able to accurately track the utilization of hosts and deploy new jobs to lesser used hosts minimizing the possibility of hot spots. I would expect a more even distribution of CPU ready time across the cluster to be representative of this.

lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +581,17 @@ def _check_pool(pool)
...
+      $redis.smembers('vmpooler__migrating__' + pool['name']).each do |vm|
+        if inventory[vm]
+          begin
+            migrate_vm(vm, pool['name'])
+          rescue
+            $logger.log('s', '[x] [' + pool['name'] + "] '" + vm + "' failed to migrate")
+          end
+        end
+      end
+
```
mattkirby: What hypervisor data are you hoping to see? The only thing that stands out to me is if you wanted to know the host your VM is running on we could report that when logging that no migration is required. This will already be reported when a migration is performed.

lib/vmpooler/vsphere_helper.rb
mattkirby: Thanks. I wasn't sure what the best practice is and have adjusted this.

lib/vmpooler/pool_manager.rb
mattkirby: @waynr, I have added this change.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,31 +196,62 @@ def find_folder(foldername)
       base
     end
 
+    def get_host_utilization(host, model=nil, limit=90)
```
mattkirby: I added some comments to try and provide some context for what these means. Let me know if you think further comments are necessary, or would be helpful.

mattkirby:
I have added an additional commit that makes VM migration at checkout optional, defaulting to disabled. When migration_limit is in the config section of vmpooler.yaml, and the value is greater than 0, migration at checkout will happen. In either case this change adds logging to vmpooler.yaml of a VM parent host, whether involving a migration or not.

shrug:
I'm 👍  on getting this in and then tuning. I have not done a deep review, but if @rick has looked it over that's good enough for me

vmpooler.yaml.example
```diff
@@ -225,6 +225,13 @@
 #     If set, prefixes all created VMs with this string.  This should include
 #     a separator.
 #     (optional; default: '')
+#
+#   - migration_limit
+#     When set to any value greateer than 0 enable VM migration at checkout.
```
rick: Typo:  "greateer"

rick:
One typo in a comment, otherwise:  LGTM. ✨ :rocket: :guitar: :metal:

waynr:
This looks pretty good to me. @mattkirby is there a particular dashboard we should watch once this change rolls out to production?

vmpooler.yaml.example
mattkirby: Thanks. Fixed it.

rick:
:+1: from me to move this forward to testing on vmpooler-dev. Could also coordinate with @puppetlabs/cinext about trying it on the cinext vmpooler. If it's looking good there, then prod vmpooler (branch deploy during time of low activity) and eventually merge.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,39 +168,98 @@ def find_folder(foldername)
...
-          if
-            (host.overallStatus == 'green') &&
-            (!host.runtime.inMaintenanceMode)
-
-            hosts[host.name] = host
-            hosts_sort[host.name] = host.vm.length
-          end
-        end
+      datacenter.hostFolder.children.each do |cluster_object|
+        return cluster_object if cluster_object.name == cluster
```
joshcooper: could shorten this to:

```ruby
datacenter.hostFolder.childen.find { |cluster_object| cluster_object.name == cluster }
```

mattkirby: Thanks. I have shortened this as suggested and updated the PR.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,39 +168,96 @@ def find_folder(foldername)
       base
     end
 
-    def find_least_used_host(cluster)
+    def get_host_utilization(host, model=nil, limit=90)
+      # Returns an array containing cumulative CPU and memory utilization of a host, and its object reference
+      # Params:
+      # +model+:: CPU arch version to match on
+      # +limit+:: Hard limit for CPU or memory utilization beyond which a host is excluded for deployments
```
nwolfe: This should docstring should go above the method signature in order for it to actually be a docstring.  Down here it's just a comment at the top of the body.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -196,39 +168,96 @@ def find_folder(foldername)
...
+      (cpu_usage.to_f / cpu_size.to_f) * 100
+    end
+
+    def memory_utilization_for(host)
+      memory_usage = host.summary.quickStats.overallMemoryUsage
+      memory_size = host.summary.hardware.memorySize / 1024 / 1024
+      (memory_usage.to_f / memory_size.to_f) * 100
+    end
+
+    def vsphere_connection_alive?(connection)
```
nwolfe: What does this method do?

At first I thought it returned a boolean value since it had the `?` in the name.

But then I saw usages didn't use it like that, so I thought it was a validation method that threw an exception.
But that doesn't appear to be the case either.

lib/vmpooler/vsphere_helper.rb
mattkirby: I have updated it. Hopefully this should look better.

lib/vmpooler/vsphere_helper.rb
nwolfe: > It verifies a connection is alive reconnects when it is not.

Ahh, I couldn't even tell that when looking at the definition itself (I had no idea what that `initialize` was going to do).

I think it would be good to reflect the behavior in the name, so something like `ensure_connected(connection)` or `reconnect_if_dead(connection)` or something like that.

nwolfe:
Is the scheduling logic sophisticated enough to warrant an english description somewhere?

Not trying to imply it is, just wondering.

It can be hard to look at the code and realize the overall behavior.

mattkirby:
Perhaps. It's hard for me to have a good sense of that from my perspective. I am happy to add some documentation if it helps make the intent more clear. It's based on calculated CPU utilization, memory utilization, then summing the two and sorting the list of compatible hosts by their usage to identify the least used.

lib/vmpooler/vsphere_helper.rb
mattkirby: I updated the name of the method. Does the updated version seem more clear? If so, I'll squash the changes.

waynr:
@puppetlabs/cinext so I was just thinking about this and wondered what everyone thinks about the possibility of having vmpooler refuse to check out a VM if it can't find a host with CPU/Memory availability below a certain threshold?

rick:
> @puppetlabs/cinext so I was just thinking about this and wondered what everyone thinks about the possibility of having vmpooler refuse to check out a VM if it can't find a host with CPU/Memory availability below a certain threshold?

If we made a change in that direction I'd prefer that it not make a vm ready rather than refusing the checkout. So the pool would be "empty" in that case.  I think it might be harder to make downstream scheduling decisions (thinking ABS here) if it looks like we have capacity but we have a higher rate of actual checkouts on that "available" capacity failing.

waynr:
Yeah my question came from ignorance (my own fault) of what's currently available in ABS.

As for refusing to make a VM ready rather than refusing a checkout, that would not prevent overutilization of the infrastructure because in that situation there is still a possibility for VMs to be ready well before overutilization begins and to cause catastrophic overutilization by being checked out willy nilly as they are currently. Anyway, setting the limit in ABS definitely suits the need that led me to ask my previous question.

mattkirby:
I've pushed a commit here that adds tests for the migration at checkout functionality. I would like to get some feedback about how the tests are written to ensure I'm testing the right pieces, and that the way the tests are structured are reasonable. I chose to separate the tests to a new file mostly just to help me isolate what I'm testing, but can merge them into a more appropriate place if it seems reasonable. I'm still working on testing the rest, but wanted to submit what I have so far here to ensure I'm on the right path.

spec/vmpooler/pool_manager_spec.rb
mattkirby: I started adding some specs here and making an effort to test things a little more thoroughly. I have not written rspec tests before, so I would like to get some feedback about how I'm approaching this to identify if I'm on the right path, or if I need to adjust the approach.

lib/vmpooler/pool_manager.rb
```diff
@@ -448,39 +460,111 @@ def _check_snapshot_queue
...
       end
     end
 
+    def find_vsphere_pool_vm(pool, vm, vsphere)
+      vsphere.find_vm(vm) || vsphere.find_vm_heavy(vm)[vm]
+    end
+
+    def migration_limit(migration_limit)
+      # Returns migration_limit setting when enabled
+      return false if migration_limit == 0 or not migration_limit
```
rick: In ruby the operator precedence of `||` vs. `or`, and `!` vs. `not` can be surprising.  I *think* this is ok here, but would prefer we use `||` and `!` instead of `or` and `not`.

lib/vmpooler/pool_manager.rb
```diff
@@ -448,39 +460,111 @@ def _check_snapshot_queue
...
+    end
+
+    def _migrate_vm(vm, pool, vsphere)
+      begin
+        $redis.srem('vmpooler__migrating__' + pool, vm)
+        vm_object = find_vsphere_pool_vm(pool, vm, vsphere)
+        parent_host, parent_host_name = get_vm_host_info(vm_object)
+        migration_limit = migration_limit $config[:config]['migration_limit']
+
+        if not migration_limit
```
rick: Ditto here for `!` preferred over `not`.

lib/vmpooler/pool_manager.rb
```diff
@@ -448,39 +460,111 @@ def _check_snapshot_queue
...
+          $logger.log('s', "[ ] [#{pool}] '#{vm}' is running on #{parent_host_name}")
+        else
+          migration_count = $redis.smembers('vmpooler__migration').size
```
rick: Wondering if `scard` would work here instead of the `smembers().size`.  Also, we might be able to fetch this at the top instead of inline, but I may not understand the intricacies here.

lib/vmpooler/pool_manager.rb
```diff
@@ -448,39 +460,111 @@ def _check_snapshot_queue
...
+          if migration_count >= migration_limit
+            $logger.log('s', "[ ] [#{pool}] '#{vm}' is running on #{parent_host_name}. No migration will be evaluated since the migration_limit has been reached")
+          else
```
rick: Purely a style preference, but I'm a fan of `return`ing at the end of an `if` branch if that's all the processing that will happen in the method.  It reduces nesting of `if`/`else` clauses and sometimes helps readability.

lib/vmpooler/pool_manager.rb
```diff
@@ -493,82 +577,96 @@ def _check_pool(pool)
       end
 
       # RUNNING
-      $redis.smembers('vmpooler__running__' + pool['name']).each do |vm|
+      $redis.smembers("vmpooler__running__#{pool['name']}").each do |vm|
         if inventory[vm]
           begin
-            check_running_vm(vm, pool['name'], $redis.hget('vmpooler__vm__' + vm, 'lifetime') || $config[:config]['vm_lifetime'] || 12)
+            check_running_vm(vm, pool['name'], $redis.hget('vmpooler__vm__' + vm, 'lifetime') || $config[:config]['vm_lifetime'] || 12, vsphere)
```
rick: Might pull the embedded `$redis.hget('vmpooler__vm__' + vm, 'lifetime') || $config[:config]['vm_lifetime'] || 12` out and assign it to a variable, for readability.

lib/vmpooler/pool_manager.rb
```diff
@@ -493,82 +577,96 @@ def _check_pool(pool)
...
         end
       end
 
       # DISCOVERED
-      $redis.smembers('vmpooler__discovered__' + pool['name']).each do |vm|
-        %w(pending ready running completed).each do |queue|
-          if $redis.sismember('vmpooler__' + queue + '__' + pool['name'], vm)
-            $logger.log('d', "[!] [#{pool['name']}] '#{vm}' found in '#{queue}', removed from 'discovered' queue")
-            $redis.srem('vmpooler__discovered__' + pool['name'], vm)
+      discovered = $redis.smembers("vmpooler__discovered__#{pool['name']}")
```
rick: Could be inlined as `if discovered = ....` instead of split across two lines.  💄

spec/helpers.rb
```diff
@@ -50,13 +50,25 @@ def create_pending_vm(template, name, token = nil)
   redis.hset("vmpooler__vm__#{name}", "template", template)
 end
 
-def create_vm(name, token = nil)
+def create_vm(name, token = nil, r = nil)
+  redis = r if r
```
rick: This is a bit tricky (I think we're shadowing a method with a local variable?).  Might be less surprising as:

``` ruby
def create_vm(name, token = nil, redis_handle = nil)
  redis_db = redis_handle ? redis_handle : redis
  redis_db.hset("vmpooler__vm__#{name}", 'checkout', Time.now)
  redis_db.hset("vmpooler__vm__#{name}", 'token:token', token) if token
end
```

spec/helpers.rb
```diff
@@ -50,13 +50,25 @@ def create_pending_vm(template, name, token = nil)
...
   redis.hset("vmpooler__vm__#{name}", 'checkout', Time.now)
   if token
     redis.hset("vmpooler__vm__#{name}", 'token:token', token)
   end
 end
 
+def create_migrating_vm(name, pool, r = nil)
+  redis = r if r
```
rick: Ditto here, with redis overriding.

spec/helpers.rb
```diff
@@ -50,13 +50,25 @@ def create_pending_vm(template, name, token = nil)
...
+  redis.hset("vmpooler__vm__#{name}", 'checkout', Time.now)
+  redis.sadd("vmpooler__migrating__#{pool}", name)
+end
+
+def add_vm_to_migration_set(name, r=nil)
+  redis = r if r
```
rick: Ditto here, with redis overriding.

spec/vmpooler/pool_manager_migration_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'spec_helper'
+require 'mock_redis'
+require 'time'
+
+describe 'Pool Manager' do
+  let(:logger) { double('logger') }
+  let(:redis) { MockRedis.new }
+  #let(:redis) { Redis.new }
```
rick: Git has your back; feel free to drop commented bits.

spec/vmpooler/pool_manager_migration_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  let(:pool) { config[:config][:pools][0]['name'] }
+  let(:vm) {
+    {
+      'name' => 'vm1',
+      'host' => 'host1',
+      'template' => pool,
+    }
+  }
+
+  before(:each) do
```
rick: This might not be necessary any more, since you're using `MockRedis`.

spec/vmpooler/pool_manager_migration_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  describe '#_migrate_vm' do
+    let(:vsphere) { double(pool) }
+    let(:pooler) { Vmpooler::PoolManager.new(config, logger, redis, metrics) }
+    context 'evaluates VM for migration and logs host' do
+      before do
+        create_migrating_vm vm['name'], pool, redis
+        allow(vsphere).to receive(:find_vm).and_return(vm)
+        allow(pooler).to receive(:get_vm_host_info).and_return([{'name' => 'host1'}, 'host1'])
```
rick: Is there any way to let the redis state (presumably set up via `create_migrating_vm`) allow us to get away from mocking this call (and similar), and just exercise the db?  Or is there too much complexity behind this?

lib/vmpooler/pool_manager.rb
mattkirby: Whoops. I'll fix it. This is some residual mess from me playing with re-arranging somethings. Thanks for catching it.

lib/vmpooler/pool_manager.rb
mattkirby: `scard` definitely works in place of `smembers().size` here. It looks like that's what I should have used, and it has now been updated.

lib/vmpooler/pool_manager.rb
mattkirby: That seems reasonable and like it helps make more clear the intent. I have added this where it looks appropriate to me in this method. It looks like I can just place `return` alone and have it return `nil`. Is that a reasonable convention, or is it preferred to return `true`?

spec/vmpooler/pool_manager_migration_spec.rb
rick: I think it's fine. I'm willing to add a bit more technical debt here (with the notion that One Day™ we'll clean up the pool manager wrt better encapsulation and testability). Just wanted to make sure that we picked the low hanging :apple: if it were there, but sounds like (not surprisingly) it's not.

lib/vmpooler/pool_manager.rb
mattkirby: I agree. Changing it.

spec/vmpooler/pool_manager_spec.rb
rick: I went through some months ago and converted most of our specs away from mocking/stubbing on redis, etc.  I stopped short on the pool_manager specs because there's just a lot of mess in that library and it needs deeper work (I think) to clean things up. 

I'm fine here with continuing in the existing style (because the cost of not doing so is large), even if the tests end up less than ideal: brittle, not fully exercising functionality.

rick:
A few comments inline, but nothing really blocking. If this has been tested to behave as desired, I'm :+1:

spec/vmpooler/pool_manager_migration_spec.rb
mattkirby: I do think some of the components I have re-arranged here will lend themselves to this being easier test. If it's valuable I'm happy to follow up on this to clean up pool manager further and implement tests around the rest of it as part of a future change.

Also, I think that these changes set us up to be able to actually test vsphere helper, which did not look very possible before.

mattkirby:
It looks like my changes to pool manager have caused the original pool manager mockist tests to fail. I will need to update these too so checks pass.

rick:
Tests passing!  🍏  💚  📗  :tea: 🇬🇱  ♻️

rick:
I'm :+1:

rick:
:+1: on the lost VM changes.

lib/vmpooler/pool_manager.rb
```diff
@@ -40,25 +39,38 @@ def _check_pending_vm(vm, pool, timeout, vsphere)
...
-
-        $redis.smove('vmpooler__pending__' + pool, 'vmpooler__completed__' + pool, vm)
+    def remove_nonexistent_vm(vm, pool)
+      $redis.srem("vmpooler__pending__#{pool}", vm)
+      $logger.log('d', "[!] [#{pool}] '#{vm}' no longer exists. Removing from pending.")
+    end
 
-        $logger.log('d', "[!] [#{pool}] '#{vm}' marked as 'failed' after #{timeout} minutes")
+    def fail_pending_vm(vm, pool, timeout, exists=true)
+      begin
```
mattkirby: That's handy. I have updated the fail_pending_vm method accordingly. Let me know if it doesn't look right.

rick:
:+1: on `rescue`-related changes.

lib/vmpooler/vsphere_helper.rb
```diff
@@ -6,14 +6,23 @@ class VsphereHelper
...
 
-    def initialize(_vInfo = {})
-      config_file = File.expand_path('vmpooler.yaml')
-      vsphere = YAML.load_file(config_file)[:vsphere]
-
-      @connection = RbVmomi::VIM.connect host: vsphere['server'],
-                                         user: vsphere['username'],
-                                         password: vsphere['password'],
-                                         insecure: true
+    def initialize(credentials)
```
mattkirby: One benefit to this change is that connections are no longer made to vsphere when a thread is initialized. Instead, they are created when work is performed. The result is the application and its threads are available faster, and we do not open as many connections as pools when the application starts.

lib/vmpooler/pool_manager.rb
```diff
@@ -20,39 +20,62 @@ def initialize(config, logger, redis, metrics)
...
-          Timeout.timeout(5) do
-            TCPSocket.new vm, 22
-          end
-          move_pending_vm_to_ready(vm, pool, host)
-        rescue
-          fail_pending_vm(vm, pool, timeout)
+          yield sock if block_given?
+        ensure
+          sock.close
+          return true
```
joshcooper: I think you'll have fewer changes if this method either yields the socket or raises an exception (so the caller doesn't need to check the return value)

lib/vmpooler/pool_manager.rb
```diff
@@ -20,39 +20,62 @@ def initialize(config, logger, redis, metrics)
...
-          (((Time.now - Time.parse(clone_stamp)) / 60) > timeout)
+      if ! host
+        fail_pending_vm(vm, pool, timeout, false)
+        return
+      end
+      socket = open_socket vm
+      fail_pending_vm(vm, pool, timeout) if ! socket
+      move_pending_vm_to_ready(vm, pool, host)
+    rescue
+      fail_pending_vm(vm, pool, timeout)
```
joshcooper: for example this could just be:

```ruby
  open_socket vm
  move_pending_vm_to_ready(vm, pool, host)
rescue
  fail_pending_vm(vm, pool, timeout)
```

lib/vmpooler/pool_manager.rb
mattkirby: You're right. I have updated it and it seems more clear this way.

mattkirby:
I think this latest commit should address the suggestions in the last few comments @joshcooper.

mattkirby:
I would like to propose merging this on 11/22 at 11am PST if there are no further issues to be raised here. Please let me know if this seems unreasonable and if additional changes are required before merging this.

I plan to rebase the changes before merging, but will hold off for tomorrow morning in order to provide reasonable time for feedback.

- [ ] Rebase commits to clean up commit history in preparation for merge.
- [ ] Wait until 11am PST to provide the opportunity for any further objections to be raised

rick:
Just to double-check: have you been able to deploy this successfully to vmpooler-dev and observe the migration?

mattkirby:
@rick, yes. The pre-tests incarnation of these changes are running on vmpooler-dev, vmpooler-cinext, and vmpooler-provisioner-dev-2 and have been running for the past several weeks. The latest commit here is running on vmpooler-provisioner-dev-2. I am happy to deploy the latest commit here to dev and vmpooler-cinext before deploying to production if that is helpful as well. My plan was otherwise to update each of those to be running against the merged commit prior to updating production vmpooler.

rick:
Sounds great to me :+1:

mattkirby:
I have rebased commits to clean up the history in preparation for merge.

PR #169, opened 2016-11-19
Containerize vmpooler

sschneid:
Built off the [jruby](https://hub.docker.com/_/jruby/) `Dockerfile`, it installs gem requirements via `bundler`, installs and runs a `redis-server`, imports the current working directory, starts `vmpooler`, and tails the logfile.

Creating a `vmpooler.yaml` configuration file in the current working directory and running the following provides a development environment:

```
docker build -t vmpooler . && docker run -p 80:4567 -it vmpooler
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,7 @@ FROM jruby:1.7-jdk
 RUN mkdir -p /var/lib/vmpooler
 WORKDIR /var/lib/vmpooler
 
-ADD Gemfile /var/lib/vmpooler
-ADD Gemfile.lock /var/lib/vmpooler
+ADD Gemfile* /var/lib/vmpooler
 RUN bundle install --system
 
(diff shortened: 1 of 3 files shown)


rick:
I dig it.  :+1:  Wonder if we can fold the `Dockerfile` build instructions into either the main README or (maybe simpler) a comment at the top of the `Dockerfile`?

shrug:
Ping @sschneid Do you want to update the instructions per @rick's comment? Otherwise we can merge this and update the docs if you'd rather pitch it over the wall.

sschneid:
@shrug Sure thing -- I'll update both the `Dockerfile` as well as add a snippet to the `README`.  Stay tuned!

sschneid:
@shrug @rick : I've added the requested documentation, as well as a small patch to allow loading configuration through an environment parameter rather than requiring an embedded configuration file within the container.

PR #170, opened 2016-11-19
Fix JavaScript error on nil `weekly_data`

sschneid:
Without this patch the dashboard will error-out on unfound `weekly_data` values.


rick:
:+1:

PR #172, opened 2016-11-22
(POOLER-44) Fix vmpooler.migrate reference

mattkirby:
This commit updates vmpooler.migrate metric to send the pool name. Without this change the VM name is sent. Additionally, counts are added for migrate_from_#host and migrate_to_#host in order to allow tracking host migrations, and correlating them with CPU ready time changes.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -413,7 +413,7 @@ module Vmpooler
     def check_disk_queue
       $logger.log('d', "[*] [disk_manager] starting worker thread")
 
-      $vsphere['disk_manager'] ||= Vmpooler::VsphereHelper.new $config[:vsphere]
+      $vsphere['disk_manager'] ||= Vmpooler::VsphereHelper.new $config[:vsphere], $metrics
 
       $threads['disk_manager'] = Thread.new do
         loop do
@@ -439,7 +439,7 @@ module Vmpooler
     def check_snapshot_queue
       $logger.log('d', "[*] [snapshot_manager] starting worker thread")
 
-      $vsphere['snapshot_manager'] ||= Vmpooler::VsphereHelper.new $config[:vsphere]
+      $vsphere['snapshot_manager'] ||= Vmpooler::VsphereHelper.new $config[:vsphere], $metrics
 
       $threads['snapshot_manager'] = Thread.new do
         loop do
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -531,11 +531,13 @@ def remove_vmpooler_migration_vm(pool, vm)
...
 
-    def migrate_vm_and_record_timing(vm_object, vm_name, host, vsphere)
+    def migrate_vm_and_record_timing(vm_object, vm_name, pool, host, source_host_name, dest_host_name, vsphere)
       start = Time.now
       vsphere.migrate_vm_host(vm_object, host)
       finish = '%.2f' % (Time.now - start)
-      $metrics.timing("migrate.#{vm_name}", finish)
+      $metrics.timing("migrate.#{pool}", finish)
+      $metrics.increment("migrate_from.#{source_host_name}")
+      $metrics.increment("migrate_to.#{dest_host_name}")
```
rick: How many host names do we have? Just want to know if we are creating lots of metrics or a fairly small set.

lib/vmpooler/pool_manager.rb
```diff
@@ -531,11 +531,13 @@ def remove_vmpooler_migration_vm(pool, vm)
       end
     end
 
-    def migrate_vm_and_record_timing(vm_object, vm_name, host, vsphere)
+    def migrate_vm_and_record_timing(vm_object, vm_name, pool, host, source_host_name, dest_host_name, vsphere)
       start = Time.now
       vsphere.migrate_vm_host(vm_object, host)
       finish = '%.2f' % (Time.now - start)
-      $metrics.timing("migrate.#{vm_name}", finish)
+      $metrics.timing("migrate.#{pool}", finish)
```
waynr: I didn't realize this was recording the migration time. I think it would be useful to also include a `$metrics.increment` for the `migrate.#{pool}` endpoint, do you know if it's possible to record both types of information at the same endpoint? Or would they require different endpoints altogether?

waynr: Actually based on looking at how we are getting `vmpooler.destroy` and `vmpooler.clone` metrics elsewhere in this source code it looks like the `$metrics.timing` values do give us `vmpooler.{clone,destroy,migrate}.<pool>.count` already so nevermind.

waynr:
:+1:

lib/vmpooler/pool_manager.rb
mattkirby: 31 total, for now, when considering the 10-node mac mini cluster, and 21-node acceptance cluster.

rick:
:cool:  :+1:

waynr:
@mattkirby this is marked as "needs testing"...have you had a chance to try it out yet?

mattkirby:
Yes. It's running now on vmpooler-provisioner-dev-2, along with the connection monitoring change so we could track new connections. It looks to be working how I would expect so far.

PR #174, opened 2016-12-01
(POOLER-48) Clear migrations at application start time

mattkirby:
This commit updates vmpooler to clear the migrations queue at application start time. When the application is shut down it is not considerate of any activities, like migrations, in flight. The result is that when the application is started again any stale entries in vmpooler__migration will be left until manually removed, which can prevent migrations from occurring.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -705,6 +705,8 @@ module Vmpooler
 
       # Clear out the tasks manager, as we don't know about any tasks at this point
       $redis.set('vmpooler__tasks__clone', 0)
+      # Clear out vmpooler__migrations since stale entries may be left after a restart
+      $redis.del('vmpooler__migration')
 
       loop do
         if ! $threads['disk_manager']


lib/vmpooler/pool_manager.rb
```diff
@@ -705,6 +705,8 @@ def execute!
 
       # Clear out the tasks manager, as we don't know about any tasks at this point
       $redis.set('vmpooler__tasks__clone', 0)
+      # Clear out vmpooler__migrations since stale entries may be left after a restart
+      $redis.del('vmpooler__migration')
```
waynr: Could you quickly explain the difference between clearing out the tasks manager and clearing out migrations? I'm mostly just curious why one has a value set to 0 to clear it out and the other is deleted completely.

mattkirby: vmpooler__tasks__clone is a counter that has its number increased, and decreased, as clone operations start and finish.

Rather than replicate that model I thought it would be preferred to know which VMs are in the process of being migrated, so I create a set for vmpooler__migration, the vm name is a member of the set, and we count the set members to determine whether we have hit the migration limit. My thought was that it would be helpful to be able to find out which VMs are in the process of being migrated if ever we need to, which I have found useful when validating that an entry is stale.

waynr:
:+1:

rick:
:+1:

PR #175, opened 2016-12-01
(POOLER-47) Send clone errors up

mattkirby:
This commit updates vmpooler to ensure clone errors, and other pool manager errors are raised to the parent method. Without this change vmpooler gets stuck after a connection fails during clone operations and will not attempt to clone again.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -628,7 +628,6 @@ module Vmpooler
           begin
             destroy_vm(vm, pool['name'], vsphere)
           rescue => err
-            $logger.log('s', "[!] [#{pool['name']}] '#{vm}' destroy appears to have failed")
             $redis.srem("vmpooler__completed__#{pool['name']}", vm)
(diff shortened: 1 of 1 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -621,11 +627,12 @@ def _check_pool(pool, vsphere)
         if inventory[vm]
           begin
             destroy_vm(vm, pool['name'], vsphere)
-          rescue
+          rescue => err
             $logger.log('s', "[!] [#{pool['name']}] '#{vm}' destroy appears to have failed")
```
rick: Might be able to consolidate this log line with the new one below.

waynr:
I'm pretty much +1 on this, for whatever that's worth given my general lack of familiarity with vmpooler code base. If the code were a little easier to work with in the first place and if there wasn't a somewhat urgent need to deploy this I might suggest writing tests to validate the logging and follow-up attempts to clone again.

PR #177, opened 2017-01-24
(maint) Enhance VM Pooler developer experience

glennsarti:
Previously, a bundle install would not pull in gems from Gemfile.local or
~/.gemfile which are common development workflows in Puppet.  This commit
modifies the Gemfile to pull in these additional gemfiles if they exist.  This
commit also adds common files and folders to gitignore which should not be
committed to this repository.

Previously, if you ran the vpooler via ruby, pressing Ctrl-C would terminate the
Webserver however the PoolManager does not have a handler and would instead
just keep executing.  This commit adds a global Ctrl-C hook which terminates
both the api and manager threads.  This behaviour will only be enabled if the
`VMPOOLER_DEBUG` environment variable exists so that it does not affect VMPooler
when running in production environments.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,3 +23,13 @@ group :test do
   gem 'simplecov', '>= 0.11.2'
   gem 'yarjuf', '>= 2.0'
 end
+
+# Evaluate Gemfile.local if it exists
+if File.exists? "#{__FILE__}.local"
+  instance_eval(File.read("#{__FILE__}.local"))
+end
+
+# Evaluate ~/.gemfile if it exists
+if File.exists?(File.join(Dir.home, '.gemfile'))
+  instance_eval(File.read(File.join(Dir.home, '.gemfile')))
+end
--- a/vmpooler
+++ b/vmpooler
@@ -26,4 +26,11 @@ manager = Thread.new {
   ).execute!
 }
 
+if ENV['VMPOOLER_DEBUG']
+  trap("INT") {
+    puts "Shutting down."
+    [api, manager].each { |t| t.exit }
+  }
+end
+
 [api, manager].each { |t| t.join }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 .ruby-version
 Gemfile.lock
+Gemfile.local
 vendor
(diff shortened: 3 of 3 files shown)


glennsarti:
Needs to fix https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler.rb#L28 as well..Should be load_file not just load

glennsarti:
Ready for review

mckern:
Some of the commits in this PR are also in #180 -- is there a specific merge order desired here?

glennsarti:
Yes.. This first then 180.  I added in the PR comments 180 that this builds on 177.  Perhaps that wasn't as obvious as it should be.

Gemfile
```diff
@@ -23,3 +23,13 @@ group :test do
...
 end
+
+# Evaluate Gemfile.local if it exists
+if File.exists? "#{__FILE__}.local"
+  eval(File.read("#{__FILE__}.local"), binding)
+end
+
+# Evaluate ~/.gemfile if it exists
+if File.exists?(File.join(Dir.home, '.gemfile'))
+  eval(File.read(File.join(Dir.home, '.gemfile')), binding)
```
mckern: Like I said in #180, I'm not generally a fan of the `Gemfile.local` pattern, but it's not something I feel strongly enough about to push back on.

BUT! You can eschew the passing of `binding` if you use `instance_eval` instead:

```
instance_eval(File.read("#{__FILE__}.local")
```

glennsarti:
@mckern Updated with `instance_eval`

vmpooler
```diff
@@ -26,4 +26,11 @@ manager = Thread.new {
   ).execute!
 }
 
+if ENV['VMPOOLER_DEBUG']
+  trap("INT") {
+    puts "Shutting down."
+    [api, manager].each { |t| t.exit }
+  }
```
mckern: Nitpick: multiline blocks should be `do`/`end`, not braces. Not gonna block on this but maybe this repo needs some Rubocop love?

glennsarti: Fixed

PR #179, opened 2017-01-26
{WIP}(POOLER-70)(POOLER-71)(POOLER-72) Refactor out VSphere for easier VM Pooler development

glennsarti:
WIP PR Only...DO not merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 .ruby-version
 Gemfile.lock
+Gemfile.local
 vendor
+vmpooler.yaml
+.bundle
+coverage
--- a/Gemfile
+++ b/Gemfile
@@ -23,3 +23,13 @@ group :test do
   gem 'simplecov', '>= 0.11.2'
   gem 'yarjuf', '>= 2.0'
 end
+
+# Evaluate Gemfile.local if it exists
+if File.exists? "#{__FILE__}.local"
+  eval(File.read("#{__FILE__}.local"), binding)
+end
+
+# Evaluate ~/.gemfile if it exists
+if File.exists?(File.join(Dir.home, '.gemfile'))
+  eval(File.read(File.join(Dir.home, '.gemfile')), binding)
+end
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -12,7 +12,7 @@ module Vmpooler
   require 'yaml'
   require 'set'
 
-  %w( api graphite logger pool_manager vsphere_helper statsd dummy_statsd ).each do |lib|
+  %w( api graphite logger pool_manager vsphere_helper statsd dummy_statsd backingservice ).each do |lib|
     begin
       require "vmpooler/#{lib}"
     rescue LoadError
@@ -21,7 +21,7 @@ module Vmpooler
   end
 
   def self.config(filepath='vmpooler.yaml')
-    parsed_config = {}
+  parsed_config = {}
 
     if ENV['VMPOOLER_CONFIG']
       # Load configuration from ENV
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -56,6 +56,8 @@ module Vmpooler
 
       def authenticate(auth, username_str, password_str)
         case auth['provider']
+          when 'dummy'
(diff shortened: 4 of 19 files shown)


glennsarti:
TLDR: It's going to be painful no matter what :-(

@rick Without a good set of tests prior to the refactor you're 100% correct that this is going to be difficult.  However due to how intertwined vSphere is, we couldn't write good tests in the first place.

I personally feel, that because we're coming from a place of fragility/unknown edge cases, no matter what we do (refactor or just leave alone) it's painful.  I feel it's better to take the pain of refactor but make sure that we more obvious expectations of the code through tests, documentation, logical design etc.

I think I need to do a few things:
1. Instead of lumping everything into a single commit (like I have done) separate out the functionality into distinct commits e.g. This commit adds the create_vm ability, this commit adds the destroy_vm etc.

2. Add a good set of spec tests on the pool_manager and dummy provider so that we have a good baseline of the expected behaviour.  Once we know how they're supposed to behave it makes it easier to reason about the behaviour of the vSphere side of things.

3.Yeah... Testing the newly refactored vSphere classes is going to be difficult.  In theory, you could setup a beaker acceptance test to spin up a vCenter instance, AD DC and ESXi hypervisor to run acceptance tests.  I think for the sake of the refactor, we need to create some mock-y spec tests and then manually test it on a separate vCenter instance (not that hard to do), but document the tests in a ticket.

rick:
> @rick Without a good set of tests prior to the refactor you're 100% correct that this is going to be difficult.  However due to how intertwined vSphere is, we couldn't write good tests in the first place.

There are some techniques in [Working Effectively with Legacy Code](https://www.amazon.com/Working-Effectively-Legacy-Michael-Feathers/dp/0131177052) that apply to these sorts of situations:  characterization testing, sensing variables, refactoring to tests, etc.  I would probably approach this step-by-step that way.  There's an ancient [video](https://www.youtube.com/watch?v=i5CG-ncx2hM) of me (I shudder when I watch it) where I show how to apply this stuff, via bringing tests to [flog](https://github.com/seattlerb/flog) for adding functionality.  

> I think I need to do a few things:
> 1. Instead of lumping everything into a single commit (like I have done) separate out the functionality into distinct commits e.g. This commit adds the create_vm ability, this commit adds the destroy_vm etc.

I agree with that approach. It makes it possible to review it commit-by-commit, and if tests are being added during extraction then it's possible to keep the refactoring green at each step.

> 2. Add a good set of spec tests on the pool_manager and dummy provider so that we have a good baseline of the expected behaviour.  Once we know how they're supposed to behave it makes it easier to reason about the behaviour of the vSphere side of things.

:+1:  I recommend those techniques from "Working Effectively..." for this sort of work.

> 3.Yeah... Testing the newly refactored vSphere classes is going to be difficult.  In theory, you could setup a beaker acceptance test to spin up a vCenter instance, AD DC and ESXi hypervisor to run acceptance tests.  I think for the sake of the refactor, we need to create some mock-y spec tests and then manually test it on a separate vCenter instance (not that hard to do), but document the tests in a ticket.

Yeah, it'll be hard no matter what, if we're working on vsphere directly. I think it would be great (and we've always wanted it) to have a few very simple high-level acceptance tests that (1) bring vsphere into a known state, (2) check out vms (partial pool, full pool, attempt more than a pool size), and (3) return them.

Otherwise, refactoring to a place where the actual vsphere back-end provider is in a class, and there is otherwise a dummy class that we can use in all non-direct tests is probably the direction to go.

This stuff is always slow (which is why I backed away, as I wasn't ready to commit those kinds of resources), but it's doable if it's done systematically.

A side note, one of the reasons characterization tests are important is that to refactor we add a scaffold of tests to assert the current behavior, and then we can do a behavior-preserving refactoring with test support. The reality is that there are bugs in vmpooler. If we try to fix the bugs as we're adding tests we will introduce more bugs.  A characterization test says, "This is the **CURRENT** behavior, not the correct behavior", letting us refactor; and then we write the test for the correct behavior, fix the bug, and drop the characterization tests.

glennsarti:
Thanks @rick.  That's some good stuff I can work with.

glennsarti:
Will raise a series of PRs for this work.  For the moment, closing this PR.

PR #180, opened 2017-02-05
(POOLER-71) Add dummy authentication provider

glennsarti:
Previously it was difficult to do local development as VMPooler requires an LDAP
service for authentication. This commit adds a dummy authentication provider.
The provider has passes authentication if the username and password are
different, and fails if the username and password are the same.  This commit
also updates the documentation in the config YML file.

### DIFF
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -32,6 +32,19 @@ module Vmpooler
       parsed_config = YAML.load_file(config_file)
     end
 
+    # Bail out if someone attempts to start vmpooler with dummy authentication
+    # without enbaling debug mode.
+    if parsed_config[:auth]['provider'] == 'dummy'
+      unless ENV['VMPOOLER_DEBUG']
+        warning = [
+          "Dummy authentication should not be used outside of debug mode",
+          "please set environment variable VMPOOLER_DEBUG to 'true' if you want to use dummy authentication",
+        ]
+
+        raise warning.join(";\s")
+      end
+    end
+
     # Set some configuration defaults
     parsed_config[:redis]             ||= {}
     parsed_config[:redis]['server']   ||= 'localhost'
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -56,6 +56,8 @@ module Vmpooler
 
       def authenticate(auth, username_str, password_str)
         case auth['provider']
+          when 'dummy'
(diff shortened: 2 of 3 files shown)


shermdog:
:+1:

glennsarti:
I can rebase onto master if it's easier.

Gemfile
```diff
@@ -23,3 +23,13 @@ group :test do
   gem 'simplecov', '>= 0.11.2'
   gem 'yarjuf', '>= 2.0'
 end
+
+# Evaluate Gemfile.local if it exists
+if File.exists? "#{__FILE__}.local"
+  eval(File.read("#{__FILE__}.local"), binding)
```
mckern: I'm not generally a fan of the `Gemfile.local` pattern, but it's not something I feel strongly enough about to push back on.

BUT! You can eschew the passing of `binding` if you use `instance_eval` instead:

```
instance_eval(File.read("#{__FILE__}.local")
```

mckern: I'll piggyback these comments over to #177

glennsarti:
@mckern Actually I can do one of a few things.  As part of #177 I added a VMPOOLER_DEBUG environment variable.

1. I can make the authentication fail if the dummy auth is used but the VMPOOLER_DEBUG env var doesn't exist, and of course log a verbose message
2. I can make the authentication proceed but log a console warning that this should be used in development
3. I can modify the API entry point (https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler.rb) where it loads the configuration, and fail to even start the API server if dummy auth is configured but the VMPOOLER_DEBUG env var is missing.

I'm leaning towards option 3.  Fail hard, fail fast.

glennsarti:
Rebased off of master now.  The PR ordering just served to confuse things.

lib/vmpooler.rb
```diff
@@ -32,6 +32,17 @@ def self.config(filepath='vmpooler.yaml')
       parsed_config = YAML.load_file(config_file)
     end
 
+    # Bail out if someone attempts to start vmpooler with dummy authentication
+    # without enbaling debug mode.
+    if parsed_config[:auth]['provider'] == 'dummy' && ENV['VMPOOLER_DEBUG']
```
mckern: No, this needs to be fixed. Good job, me!

glennsarti:
Changes are fine by me.  LGTM

PR #187, opened 2017-02-22
(maint) Only load rubocop rake tasks if gem is available

glennsarti:
Previously in commit 145ecb443f Rakefile was modified to require the Rubocop
gem however, this gem is not available on jruby systems and would throw errors.
This commit modifies the Rakefile to first detect if the Gem is available and
optionally load the rubocop rake tasks.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,7 @@
 require 'rspec/core/rake_task'
-require 'rubocop/rake_task'
+
+rubocop_available = Gem::Specification::find_all_by_name('rubocop').any?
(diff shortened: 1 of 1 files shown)


mckern:
Should we add a corresponding confinement for the Rubocop gem in Gemfile?

glennsarti:
The Gem is already confined to the MRI and Windows x64 platforms which is what caused the Travis failures in the first place (jruby was failing)

mckern:
works for giant man

PR #189, opened 2017-03-02
(POOLER-70) Refactor clone_vm to take pool configuration object

glennsarti:
Previously, the clone_vm method took various VSphere specific parameters e.g.
template folder.  However in order make VMPooler less VSphere specific this
method should just take the pool configuration and then it can determine the
appropriate settings itself.  This commit also moves the threading to a clone_vm
while the actual method which does the work is now _clone_vm as per all other
multithread worker methods in pool_manager.  This commit also updates the spec
tests appropriately.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -185,103 +185,111 @@ module Vmpooler
     end
 
     # Clone a VM
-    def clone_vm(template, folder, datastore, target, vsphere)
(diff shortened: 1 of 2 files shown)


glennsarti:
The diff for this looks strange because `clone_vm` has been re purposed.  All of the worker code was moved to `_clone_vm`

underscorgan:
This diff is much easier with `?w=1` :D, looks good!

PR #190, opened 2017-03-03
(POOLER-70) Prepare to refactor VSphere code into a VM Provider

glennsarti:
Previously all of the VM provisioning code was intertwined with the VM lifecycle
code e.g. The VSphere specific code is mixed with Redis code.  This makes it
impossible to add aditional providers or disable VSphere integration.  This
commit begins the process to refactor the VSphere code out of the lifecycle code
by introducing the concept of VM Providers.  A Provider will contain the logic/
code to manage VMs i.e. create/destroy/inquire.  Therefore the Pool Manager can
query a strict interface into one or more Providers.  Initially only a VSphere
provider will be available.

This commit adds the base class for all providers and describes the API or
contract that the Pool Manager will use to manage VMs.

VM provisioning will be handled by VM Providers.  This commit renames the use of
vsphere to provider where appropriate and changes the per-pool helper from
vsphere to providers to more accurately represent it's intended use.

This commit adds a skeleton VM Provider for VSphere based VM operations.

### DIFF
--- a/lib/vmpooler/providers.rb
+++ b/lib/vmpooler/providers.rb
@@ -1,0 +1,7 @@
+%w( base vsphere ).each do |lib|
+  begin
+    require "vmpooler/providers/#{lib}"
+  rescue LoadError
+    require File.expand_path(File.join(File.dirname(__FILE__), 'providers', lib))
+  end
+end
--- a/lib/vmpooler/providers/base.rb
+++ b/lib/vmpooler/providers/base.rb
@@ -1,0 +1,107 @@
+module Vmpooler
+  class PoolManager
+    class Provider
+      class Base
+        # These defs must be overidden in child classes
+
+        def initialize(options)
+          @provider_options = options
+        end
+
+        # returns
+        #   [String] Name of the provider service
+        def name
+          'base'
+        end
+
+        # inputs
+        #  pool : hashtable from config file
+        # returns
+        #   hashtable
+        #     name : name of the device   <---- TODO is this all?
+        def vms_in_pool(pool)
+          fail "#{self.class.name} does not implement vms_in_pool"
+        end
+
+        # inputs
+        #   vm_name: string
+        # returns
+        #    [String] hostname   = Name of the host computer running the vm.  If this is not a Virtual Machine, it returns the vm_name
+        def get_vm_host(vm_name)
+          fail "#{self.class.name} does not implement get_vm_host"
+        end
+
+        # inputs
+        #   vm_name: string
+        # returns
+        #    [String] hostname   = Name of the most appropriate host computer to run this VM.  Useful for load balancing VMs in a cluster
+        #                          If this is not a Virtual Machine, it returns the vm_name
+        def find_least_used_compatible_host(vm_name)
+          fail "#{self.class.name} does not implement find_least_used_compatible_host"
+        end
+
+        # inputs
+        #   vm_name: string
+        #   dest_host_name: string (Name of the host to migrate `vm_name` to)
+        # returns
+        #    [Boolean] Returns true on success or false on failure
+        def migrate_vm_to_host(vm_name, dest_host_name)
+          fail "#{self.class.name} does not implement migrate_vm_to_host"
+        end
+
+        # inputs
+        #   vm_name: string
+        # returns
+        #   nil if it doesn't exist
+        #   Hastable of the VM
+        #    [String] name       = Name of the VM
+        #    [String] hostname   = Name reported by Vmware tools (host.summary.guest.hostName)
+        #    [String] template   = This is the name of template exposed by the API.  It must _match_ the poolname
+        #    [String] poolname   = Name of the pool the VM is located
+        #    [Time]   boottime   = Time when the VM was created/booted
+        #    [String] powerstate = Current power state of a VM.  Valid values (as per vCenter API)
+        #                            - 'PoweredOn','PoweredOff'
+        def get_vm(vm_name)
+          fail "#{self.class.name} does not implement get_vm"
+        end
+
+        # inputs
+        #   pool : hashtable from config file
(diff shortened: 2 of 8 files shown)


spec/unit/providers/base_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(subject.vm_exists?('vm')).to eq(true)
+    end
+
+    it 'should return false when get_vm returns nil' do
+      allow(subject).to receive(:get_vm).with('vm').and_return(nil)
+
+      expect(subject.vm_exists?('vm')).to eq(false)
+    end
+  end
+end
```
underscorgan: missing newline

spec/unit/providers/vsphere.rb
```diff
@@ -0,0 +1,92 @@
...
+      expect(subject.vm_exists?('vm')).to eq(true)
+    end
+
+    it 'should return false when get_vm returns nil' do
+      allow(subject).to receive(:get_vm).with('vm').and_return(nil)
+
+      expect(subject.vm_exists?('vm')).to eq(false)
+    end
+  end
+end
```
underscorgan: missing newline

lib/vmpooler/providers/vsphere.rb
```diff
@@ -0,0 +1,17 @@
...
+        end
+
+        def name
+          'vsphere'
+        end
+
+      end
+    end
+  end
+end
```
underscorgan: missing newline

lib/vmpooler/providers/base.rb
```diff
@@ -0,0 +1,107 @@
...
+        # returns
+        #   result: boolean
+        def vm_exists?(vm)
+          !get_vm(vm).nil?
+        end
+
+      end
+    end
+  end
+end
```
underscorgan: missing newline

lib/vmpooler/providers.rb
```diff
@@ -0,0 +1,7 @@
+%w( base vsphere ).each do |lib|
+  begin
+    require "vmpooler/providers/#{lib}"
+  rescue LoadError
+    require File.expand_path(File.join(File.dirname(__FILE__), 'providers', lib))
+  end
+end
```
underscorgan: missing newline

glennsarti:
@underscorgan Hrmmm this is odd.  I've added trailing newlines where I missed them (Thanks!) and my local diff showed it. But github is not showing the trailing newline.  Odd.

underscorgan:
@glennsarti weirdddd.

glennsarti:
Did you want to try checking out the branch on your local computer to make sure those changes stuck?

joshcooper:
@glennsarti so we're currently pinned to 8286ec2854bae2c02cdab9da1e7cdb40bd821531 in production (as we had to rollback to the pre-migration known good state):

```
/var/lib/vmpooler$ git rev-parse HEAD
8286ec2854bae2c02cdab9da1e7cdb40bd821531
```

AFAIK the difference between 8286ec28 and head of master (circa Feb or so) were not reverted. So we are stuck in a bad place, were we cannot roll out the changes you're working on until we revert the migration logic. Can you look into reverting those changes, perhaps pairing with @mattkirby on that? /cc @puppetlabs/cinext

glennsarti:
How about as part of this work on POOLER-70...the migration logic is feature flagged?

glennsarti:
FYI git compare  https://github.com/puppetlabs/vmpooler/compare/8286ec2854bae2c02cdab9da1e7cdb40bd821531...2bf7c4a7150c09ffb4b01d925f273ba1d95220fc

glennsarti:
@joshcooper Although looking at https://tickets.puppetlabs.com/browse/POOLER-50  How about I just rip it all out?

glennsarti:
@joshcooper @mattkirby  I've done some more digging and I'm leaning to keeping the migration code in.  TLDR version; This PR is good to go and will one of many PRs for the POOLER-70 work. 

Reposting my thoughts from POOLER-70

---

I thought I would document my plans for what and how to achieve the refactoring:

* The refactor would move the vSphere specific code out of the pool_manager into a VM Provider class

* The refactor would initially retain the existing behavior of one instance of the vSphere provider per pool.  Ultimately the goal is to have only one VM Provider instance for all pools, but having this intermediate step makes it easier to transition such a large code change.  

* The refactor would initially just reference the code in the VSphere helper class and then slowly move the code from the helper into the VM Provider proper.  This is intermediate step is used to make it easier transition such a large code change.

* This refactor should not black any work on improving the vSphere helper or Pool Manager itself.  Indeed it is expected that as part of the refactor errors in the logic in both the helper and pool manager will be uncovered and fixed in other issue tickets and PRs.

* Ideally this refactor should make as few behaviour changes to the Pool Manager as possible.  It is hoped that the only code changes will be due to making the code easier to test.

---

As for the overloading vSphere with connections, I will do that work as part of POOLER-50.  I've a quick review over the vSphere helper and I can see some threading issues which need to be sovled as well.

shrug:
It looks like this also needs a rebase

@joshcooper or @mattkirby Do you have any additional comments?

glennsarti:
Rebased

glennsarti:
@mattkirby The pattern I'm aiming for is similar to how Puppt handles types and providers.

The `type` is a Virtual Machine interface which the Pool Manager can call.  The ` provider` is one or more Virtual Machine management systems (e.g. vSphere, AWS, Azure, OpenStack etc., or even a Dummy one).  because the implementation of VM Provider is separate from the Pool Manager, it makes it much easier to split the logic, test it and bring in new technologies.

mattkirby:
Perhaps I should have described the ruby bits, the pattern as you described seems reasonable to me and is certainly an improvement over assuming vSphere everywhere. I think this is worth merging now and we can follow up and make further changes if need be.

PR #192, opened 2017-03-04
(WIP) (POOLER-50) Remove VM migration code

glennsarti:
(WIP)

The migrate_vm code introduced in commit 705e5d26d87a3e29a9483bf7 caused errors
and has since been found to be problematic.  This commit removes the changes
introduced as part of PR 167 that peratin to the migrate_vm code.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -55,7 +55,6 @@ module Vmpooler
 
     def account_for_starting_vm(template, vm)
       backend.sadd('vmpooler__running__' + template, vm)
-      backend.sadd('vmpooler__migrating__' + template, vm)
       backend.hset('vmpooler__active__' + template, vm, Time.now)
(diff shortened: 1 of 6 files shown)


mattkirby:
I think the problematic code is around connections in vsphere_helper, not covered here. I think the rest is functional, although its implementation becomes problematic in a pooler instance like our large production one. It is configurable and requires explicitly turning it on. I do think there is value in the concept of migration at checkout given that VMs sit for a while. Some improvements around how the connections are managed for this, and host selection, would make it less taxing in heavier usage environments.

The host selection stuff is used for host selection when provisioning VMs. The previous method counted the number of VMs and didn't consider what's running. There could be some improvements here, or we could revert to the previous method. Something would have to replace it if ripping it out.

If removing this we would want to bring something back to address POOLER-26.

glennsarti:
Good points. Closing PR.

Issue #194, opened 2017-03-05
Remove the use of global variables in the vSphere helper


glennsarti:
Currently in https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/vsphere_helper.rb#L10-L12 three global variables are used to hold information, however they are only used in the scope of the helper instance.

These variables should all be instance variables (`@xxxxxx`) as globals should be avoided in classes.

PR #196, opened 2017-03-06
(POOLER-73) Add spec tests for vsphere_helper

glennsarti:
Previously there were no spec tests to document the behaviour of the
vsphere_helper.  This commit adds the behavioural spec tests in prepration of
the refactoring work as part of POOLER-70, the connection pooling work in
POOLER-52 and various bugs found in vsphere_helper while these spec tests were
being created.

- [x] #initialize
- [x] #ensure_connected
- [x] #connect_to_vsphere
- [x] #add_disk
- [x] #find_datastore
- [x] #find_device
- [x] #find_disk_controller
- [x] #find_disk_devices
- [x] #find_disk_unit_number
- [x] #find_folder
- [x] #get_host_utilization
- [x] #host_has_cpu_model?
- [x] #get_host_cpu_arch_version
- [x] #cpu_utilization_for
- [x] #memory_utilization_for
- [x] #find_least_used_host
- [x] #find_cluster
- [x] #get_cluster_host_utilization
- [x] #find_least_used_compatible_host
- [x] #find_pool
- [x] #find_snapshot
- [x] #find_vm
- [x] #find_vm_light
- [x] #find_vm_heavy
- [x] #find_vmdks
- [x] #get_base_vm_container_from
- [x] #get_snapshot_list
- [x] #migrate_vm_host
- [x] #close

PR #198, opened 2017-03-06
Update usage of global variablesin vsphere_helper

mattkirby:
This commit updates usage of global variables in vsphere_helper to be instance variables. They do not need to be global variables as brought up in issue #194. Without this change we are setting global variables when they are not needed.

### DIFF
--- a/lib/vmpooler/vsphere_helper.rb
+++ b/lib/vmpooler/vsphere_helper.rb
@@ -7,9 +7,9 @@ module Vmpooler
     DISK_MODE = 'persistent'
 
     def initialize(config, metrics)
-      $credentials = config[:vsphere]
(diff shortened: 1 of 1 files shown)


mattkirby:
Would it be better to submit this as a PR to #196 so there's a place for them?

glennsarti:
Um.. Not sure.  My intent in 196 was to document the existing behaviour only and then we can submit later PRs to modify the behaviour.  However I'm not averse to this PR being merged prior to 196.  All it's doing is just shifting when the work gets done :-)

shrug:
I think this needs a rebase

glennsarti:
@shrug Rebased on @mattkirby 's behalf

PR #200, opened 2017-03-09
(maint) Fix credentials in vsphere_helper

glennsarti:
In commit 03ad7bfa468 the global variables for credentials was change to an
instance variable however one reference was missed.  This commit fixes this
omission.


glennsarti:
ping @mattkirby @shrug

PR #216, opened 2017-04-05
(POOLER-72)(POOLER-70)(POOLER-52) Move Pool Manager to use the VM Provider

glennsarti:
This Pull Request, unfortunately, is a large piece of work. It is composed of three parts:

1. (POOLER-70) Use a VM provider to abstract all VM specific tasks into the provider and leave the pool_manager to manage the VM pools instead.

Previously a vSphere VM provider had been worked on, but this PR is about stripping out any VM (in particular vSphere) tasks from the pool_manager, and only interface through the VM Provider.  This work was performed with a single commit for each method modified in pool_manager, to make PR review easier.

This PR also moves the configuration information from the root of the YAML into `:providers:/:<provider name>` and the associated documentation.

2. (POOLER-72) Create a Dummy VM provider to make it easier to test/develop

This PR adds a Dummy VM provider which acts like a physical VM provider, however it can be used entirely locally.  You can also alter a lot of its behaviour and inject failure and delays to see
what would happen.  Eventually this provider will be used in the integration tests to make it easier to test API changes.

3. (POOLER-52) Use a connection pooler for connecting to vSphere

Unfortunately this work had to be included in the PR.  When implementing POOLER-70 this left a single vSphere connection (one per provider) to manage ALL the vSphere pools.  This would not work when there are a medium number (> 10) of pools or high VM turnover.  The commits with this first add a generic connection pooler, which was extended to add more instrumentation for metrics.  Then the Dummy provider was modified to use the connection pooler to allow better testing, and then finally the vSphere Provider was modified to use a connection pooler with a backward compatible pool size of the number of pools.

- [x] Update documentation with new vsphere configuration location


glennsarti:
@mattkirby This is now functionality complete.  Just need to cleanup all of the commit messages and add documentation.

glennsarti:
@mattkirby This is now ready for review.

This whole thing was tested manually on our real vSphere instance with 2 pools of 15 VMs each.  Also as part of https://github.com/puppetlabs/vmpooler/issues/218 it was verified it worked in their instance.

README.md
```diff
@@ -47,12 +48,14 @@ The following YAML configuration sets up two pools, `debian-7-i386` and `debian-
     pool: 'Pooled VMs/debian-7-i386'
     datastore: 'vmstorage'
     size: 5
+    provider: vsphere
```
shrug: will this be a required addition to the configs, or is there a default?

glennsarti: No.  I've made the pool_manager default to vsphere if none is set.

fuero:
Vmpooler stops working when used sparingly as vsphere connections time out. The connection pool doesn't validate the connection it returns and thus doesn't throw out dead connections and doesn't create new ones.

glennsarti:
@fuero Ahh good point.  The connection pooler does not manage that.  Need to add a helper method back in for that.  Thanks!

glennsarti:
@fuero Updated it to restore broken connections.

glennsarti:
Spec tests are failing due to recent rack gem update.  Fix is up at https://github.com/puppetlabs/vmpooler/pull/219.

glennsarti:
Rebased.

cdenneen:
Glenn what hypervisor does the dummy use locally? 
Any chance to expand this to leverage aws, gce, etc? Spot instances might be good use case ?

glennsarti:
@cdenneen 
> Glenn what hypervisor does the dummy use locally?

None.  It's purely a in-memory hashtable to emulate a VM provider.  It can optionally output to a file.  It's mainly designed for testing VM pooler functions or just a sanity check.

glennsarti:
As for expanding.  Yes this quite possible.  This PR was mainly about extracting out the vSphere 'stuff' to facilitate other work.  If we could use a plugin system - Even Better!!!

PR #217, opened 2017-04-12
(POOLER-52) Limit vsphere connection use for checking pools

mattkirby:
This change adds a capability to limit vsphere used by pool_manager when checking pools to the number specified in task_limit. Without this change a vsphere connection is created per pool.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -601,7 +601,7 @@ module Vmpooler
 
       $providers[slot] ||= Vmpooler::VsphereHelper.new $config, $metrics
 
-      $threads[slot] = Thread.new do
+      $threads[pool['name']] = Thread.new do
         _check_pool(pool, $providers[slot])
       end
     rescue => err
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -26,6 +26,11 @@ module Vmpooler
       end
     end
 
+    def slots_available?(slots, max_slots)
+      # Returns slots in use unless max is reached
+      return slots unless slots >= max_slots
+    end
+
     def open_socket(host, domain=nil, timeout=5, port=22, &block)
       Timeout.timeout(timeout) do
         target_host = host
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -26,6 +26,11 @@ def check_pending_vm(vm, pool, timeout, provider)
       end
     end
 
+    def slots_available?(slots, max_slots)
```
glennsarti: Typically a method that ends in `?` is boolean.  Seems strange to return an object or nil.  Perhaps drop the question mark?

lib/vmpooler/pool_manager.rb
```diff
@@ -569,23 +574,38 @@ def migrate_vm_and_record_timing(vm_object, vm_name, pool, host, source_host_nam
...
-    def check_pool(pool, maxloop = 0, loop_delay = 5)
-      $logger.log('d', "[*] [#{pool['name']}] starting worker thread")
+    def evaluate_pool(pool, slots, maxloop = 0, loop_delay = 1)
+      return if $redis.hget("vmpooler__pool__#{pool['name']}", 'slot')
+      checking_pools = $redis.smembers('vmpooler__check__pool__pending')
+      return if checking_pools.include? pool['name']
+      $redis.sadd('vmpooler__check__pool__pending', pool['name'])
+      loop_count = 1
+      while (slots_available?($redis.smembers('vmpooler__check__pool').count, slots.to_i).nil?)
+        sleep(loop_delay * 3)
```
glennsarti: Should this be `* 3` ? The variable name kind of implies that it should it should delay for that many seconds, not three times that many seconds.

glennsarti:
This probably won't be thread safe under JRuby.

Instead of pushing the concurrency into Redis, could we use a simple mutex synchronised counter?

glennsarti:
I've also started adding a generic connection pooler object in https://github.com/puppetlabs/vmpooler/pull/216.  It's still work in progress.  The Dummy provider now uses it and had it working there.  I just need to modify the vsphere provider to use it which I'm currently working on.

Current commit is
https://github.com/puppetlabs/vmpooler/pull/216/commits/54abf01fe0545debee51114c964f74951ad8cb4c

lib/vmpooler/pool_manager.rb
mattkirby: No, probably not. I've removed the `* 3`.

lib/vmpooler/pool_manager.rb
mattkirby: Question mark dropped.

mattkirby:
Yes, your suggestion would probably make more sense than redis for this. I can make that change here. Another thought, I don't necessarily care about restraining the threads so much as the provider connections, so it may not make sense to manage them together.

My main goal with this is to tame the approach in pool manager to be a little more considerate of provider connections. This is a big problem in our production instances because there are so many pools, and in turn so many provider connections, that vcenter slows down. Something that restrains these connections will allow us to start running the latest version again, which is what prompted me to poke at this.

glennsarti:
I agree that we should probably be running one thread per pool (although that could be reviewed later). However IMHO the providers should be responsible for not DDOSing their upstream system, not the PoolManager.

So I feel all this code change really belongs in the vSphere Helper/Provider, which is what I've been working on for POOLER-70.

mattkirby:
Do you think the changes you're referencing will land soon enough that restraining pool_manager like this is an unnecessary intermediary step?

lib/vmpooler/pool_manager.rb
```diff
@@ -761,12 +799,7 @@ def execute!(maxloop = 0, loop_delay = 1)
         end
 
         $config[:pools].each do |pool|
-          if ! $threads[pool['name']]
-            check_pool(pool)
-          elsif ! $threads[pool['name']].alive?
-            $logger.log('d', "[!] [#{pool['name']}] worker thread died, restarting")
-            check_pool(pool)
-          end
+          evaluate_pool(pool, task_limit, maxloop, loop_delay)
```
glennsarti: This pattern will tend to favour servicing pools at the lower end of the array if a `max_loop` is set and quite possibly, higher up pools will never get serviced.

e.g. Consider a 10 pool config (P01 to P10) with task_limit 5

1st run - P01 to P05 will get a slot
     P06-P10 will silently return

P01 - P05 will return quickly (duration less than the 5 * `loop_delay`)

2nd run - P01 to P05 will get a slot
     P06-P10 will silently return

and so on.
---
However if `max_loop` of 0 is set, this turns the pool evaluation into a serial process instead of the current parallel process so VM operations will be delayed.

mattkirby: Yes. My perception was that maxloop was there only for testing purposes, and blocking further pool evaluations until a slot is available seems desirable. The current pool manager is obscenely resource intensive on the host system.

glennsarti:
In the next two weeks I think so.  I've proved the connection pooler and instrumentation works on the Dummy Provider.  It's just a big PR which requires a lot of review.  Happy to talk offline about this.

lib/vmpooler/pool_manager.rb
glennsarti: I think the first thing to do is reduce the vSphere connection count and then we can look at a Thread Pooler to properly handle queuing pool evaluations in a more sane manner. e.g. use min-max thread pool so we burst if multiple pools need something but use minimal resources when it's just doing nothing.

PR #219, opened 2017-05-12
(maint) Pin rack to 1.x

glennsarti:
Rack was updated recently and 2.x requires ruby 2.2 or above. As VMPooler should
work in older jruby, we need to be Ruby 1.9.3 compatible.  This commit pins
the rack gem to 1.x.


underscorgan:
seems reasonable

PR #222, opened 2017-06-12
Added Vagrant setup and fixed the Dockerfile so it actually works

genebean:


stahnma:
Why would we I want docker in vagrant vs just using docker? (I simply don't understand).

genebean:
@stahnma I don't like the mess Docker can leave on my laptop so I generally run all my Docker stuff inside a VM. When I am finished or "think" I am finished with something Docker-related I will generally blow away the vm and bring it back up to see if I got it right.

stahnma:
Ha, I feel the same way about Vagrant. :)

genebean:
LOL. At least to clean up vagrant all I have to do is delete a folder or two... I always end up going back to Google to remember how to clean up after Docker.

stahnma:
I always just had vbox crash my laptop. Either way, carry on.

sbeaulie:
hey why not!?

Dockerfile
```diff
@@ -8,24 +8,25 @@
...
-FROM jruby:1.7-jdk
+FROM jruby:9.1-jdk
+
+RUN mkdir -p /var/lib/vmpooler && mkdir /var/log/vmpooler
 
-RUN mkdir -p /var/lib/vmpooler
 WORKDIR /var/lib/vmpooler
 
-ADD Gemfile* /var/lib/vmpooler
+ADD Gemfile* /var/lib/vmpooler/
```
sbeaulie: this * here failed for me when I had a Gemfile + a Gemfile.lock
I don't know why there is a * wildcard here?

PR #223, opened 2017-06-15
(POOLER-83) Add ability to specify a datacenter for vsphere

glennsarti:
This PR builds on PR #224 

Previously the vsphere provider assumed that there was one and only one
datacenter (DC) in the vsphere instance.  However this is simply not true for
many vSphere installations.  This commit:
- Adds the ability to define a vSphere datacenter at the Pool or Provider level
  whereby the Pool setting takes precedence
- If no datacenter is specified the default behaviour of picking the first DC
  in the vSphere instance
- Updated all tests for the new setting
- Update the vmpooler configuration file example with relevant setting name
  and expected behaviour
- Fixed a bug in the rvmomi_helper whereby if no DC was found it would return
  all DCs.  This is opposite behaviour of the real RBVMOMI library as it returns
  nil

### DIFF
--- a/lib/vmpooler/providers/vsphere.rb
+++ b/lib/vmpooler/providers/vsphere.rb
@@ -510,6 +510,12 @@ module Vmpooler
           available_unit_numbers.sort[0]
         end
 
+        # Finds the first reference to and returns the folder object for a foldername and an optional datacenter
+        # Params:
+        # +foldername+:: the folder to find (optionally with / in which case the foldername will be split and each element searched for)
+        # +connection+:: the vsphere connection object
+        # +datacentername+:: the datacenter where the folder resides, or nil to return the first datacenter found
+        # returns a ManagedObjectReference for the first folder found or nil if none found
         def find_folder(foldername, connection, datacentername)
           datacenter = connection.serviceInstance.find_datacenter(datacentername)
           raise("Datacenter #{datacentername} does not exist") if datacenter.nil?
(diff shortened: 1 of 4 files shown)


glennsarti:
Code has been completed.  Just pending an actual test on vsphere infrastructure to make sure it all works.

glennsarti:
Verified manually, against a real vSphere instance, that the datacenter setting is being used:

* Not putting in a DC found the first DC
* Specifying a DC name that does not exist in either the pool or provider config caused vSphere to throw errors saying it did not exist
* Specifying the DC name that does exist, vmpooler ran correctly. e.g.
```yaml
:vsphere:
  ...
  datacenter: 'pdx'
...
```

glennsarti:
This PR is ready for merge

lib/vmpooler/providers/vsphere.rb
```diff
@@ -422,8 +434,9 @@ def add_disk(vm, size, datastore, connection)
           true
         end
 
-        def find_datastore(datastorename, connection)
-          datacenter = connection.serviceInstance.find_datacenter
```
sbeaulie: Would this one need changed too?
https://github.com/glennsarti/vmpooler/blob/55d286f8bb3b60c9798546b231096ea6d0005854/lib/vmpooler/providers/vsphere.rb#L427

glennsarti: Yes you're right.  I'm concerned the specs didn't pick that up.

glennsarti: Updated.  And modified the specs to fail if it regresses

shrug:
Looks like this needs a rebase

glennsarti:
There's always something :-)

glennsarti:
Rebased.

Issue #226, opened 2017-06-23
Use a dynamic check_pool period


glennsarti:
Each pool is polled on a schedule to check whether there are any tasks to perform, for example, provision new VMs to fill a pool, or destroy VMs which are no longer required. By default this value is every 5 seconds.  However, with a large number of pools, this can cause the provider to be issuing many, many requests which can cause performance problems, for example, vSphere recommends no more than 100 active connections per vCenter instance. But if you increase the check interval to a large number, then vmpooler will appear to be slow to perform tasks.   There should be settings that can be used to tune how often the provider is polling for changes to a pool.

This can be done by increasing the polling period when the pool is stable or when unimportant tasks are being performed.  The polling period is decreased back to the minimum when important tasks happen (newly discovered VMs, creating new VMs and checking for VMs that have completed cloning but not ready).  This means the pools are checked appropriately during important events, but as the pool stabilises it does not need to be checked as often.

glennsarti:
Fixed by merge

PR #227, opened 2017-06-23
(GH-226) Use a dynamic pool_check loop period

glennsarti:
Previously the check_pool would always check the pool every 5 seconds, however
with a large number of pools, this can cause resource issues inside the
providers.  This commit:
- Introduces a dynamic check_pool period which increases during stability and
  decreases when the pool is being change in an important way
- Surfaces the settings as global config defaults but can also be set on a per
  pool basis
- Adds defaults to have default of 5 to 60 seconds and decay of 2.0
- Unit tests for the new behaviour

### DIFF
--- a/spec/unit/pool_manager_spec.rb
+++ b/spec/unit/pool_manager_spec.rb
@@ -1912,18 +1912,61 @@ EOT
     end
 
     context 'with dead pool thread' do
-      let(:pool_thread) { double('thread', :alive? => false) }
-      before(:each) do
-        # Reset the global variable - Note this is a code smell
-        $threads = {}
-        $threads[pool] = pool_thread
+      context 'without check_loop_delay_xxx settings' do
+        let(:pool_thread) { double('thread', :alive? => false) }
+        let(:default_check_loop_delay_min) { 5 }
+        let(:default_check_loop_delay_max) { 60 }
+        let(:default_check_loop_delay_decay) { 2.0 }
+        before(:each) do
+          # Reset the global variable - Note this is a code smell
+          $threads = {}
+          $threads[pool] = pool_thread
+        end
+
+        it 'should run the check_pool method and log a message' do
+          expect(subject).to receive(:check_pool).with(a_pool_with_name_of(pool),
+                                                       default_check_loop_delay_min,
+                                                       default_check_loop_delay_max,
+                                                       default_check_loop_delay_decay)
+          expect(logger).to receive(:log).with('d', "[!] [#{pool}] worker thread died, restarting")
+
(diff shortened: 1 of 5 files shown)


glennsarti:
Interested in your opinion on this @mattkirby

glennsarti:
As a side note, should the defaults be a more "sane" then e.g.  Default polling period is  (num of pools) / 4 so 20 pools = 5 sec polling period.

mattkirby:
Yeah, I think it's definitely an improvement.

glennsarti:
So my thinking was restricting the number of available connections doesn't stop how often a provider is polling, it just caps the maximum rate.  So it'll reach that hard limit fairly quickly when the number of connections is far lower than pools (as we've seen).  So I figured the one metric that has the greatest effect on provider load (at least in the vSphere instance) is how often vSphere is queried for the list of VMs in a Pool.

Another option would be to split operations that do and do not require an inventory scan.  This means we can limit the most "expensive" operation (inventory) to when we _actually_ need it e.g.
Inventory is needed:
- On the first pool scan for an initial inventory
- Just after a clone_vm (new vm will appear in the inventory)
- Just after a destroy_vm

In fact that ^^ may be a better way of doing it.  As my rolling timeout is still a little naive.

mattkirby:
@glennsarti as it is I think this would be useful. In starting to work with this code with some of our vmpooler instances I find that I frequently hit the 60 second timeout. Is there anything I can do to help this one land? I'm all for further iteration, but I think this is useful as it stands now.

glennsarti:
@mattkirby Sorry, been busy fixing other bits of infra.  I think I should change this to use more sane defaults.

kevpl:
@glennsarti since you're already going to be changing the defaults, it seems better to make them into constants as well since you use them enough..

glennsarti:
Okay, I've changed the defaults to 5 -> 60 and decay of 2.0.  Also used constants and updated docs.  Added some rubocop minor fixes for good measure.

glennsarti:
Ready for merge.

glennsarti:
... and tests have failed...

glennsarti:
And tests are now good.  Ready for review agan

spec/unit/pool_manager_spec.rb
```diff
@@ -2200,6 +2384,12 @@
         subject._check_pool(pool_object,provider)
       end
 
+      it 'should return the number of discoverd of VMs' do
```
shrug: nitpick: typo in "discovered"

glennsarti: Fixed

PR #229, opened 2017-07-11
(maint) Remove phantom VMs that are in Redis but don't exist in provider

glennsarti:
Previously, if inventory failed for some reason, it would return an incomplete
set of VMs which could then cause the pool to perform off behaviours such as
fill the pool high than it should, or remove VMs which exist.  Also, if the
redis cache of VMs in a pool had a VM but it did not actually exist in the
inventory it would never be removed.

This commit:
- Immediately exits the check_pool if an error occurs during inventory
  collection
- Will mark a VM as completed if it exists in Redis, but does not exist in
  inventory
- Adds tests for these behaviours

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -603,6 +603,7 @@ module Vmpooler
         end
       rescue => err
         $logger.log('s', "[!] [#{pool['name']}] _check_pool failed with an error while inspecting inventory: #{err}")
+        return pool_check_response
       end
 
       # RUNNING
@@ -615,6 +616,8 @@ module Vmpooler
           rescue => err
             $logger.log('d', "[!] [#{pool['name']}] _check_pool with an error while evaluating running VMs: #{err}")
           end
+        else
+          move_vm_queue(pool['name'], vm, 'running', 'completed', 'is a running VM but is missing from inventory.  Marking as completed.')
         end
       end
 
@@ -627,6 +630,8 @@ module Vmpooler
           rescue => err
             $logger.log('d', "[!] [#{pool['name']}] _check_pool failed with an error while evaluating ready VMs: #{err}")
           end
+        else
+          move_vm_queue(pool['name'], vm, 'ready', 'completed', 'is a ready VM but is missing from inventory.  Marking as completed.')
         end
       end
 
--- a/spec/unit/pool_manager_spec.rb
+++ b/spec/unit/pool_manager_spec.rb
@@ -2347,12 +2347,13 @@ EOT
   task_limit: 10
 :pools:
   - name: #{pool}
-    size: 0
+    size: 10
 EOT
       )
     }
     let(:pool_object) { config[:pools][0] }
     let(:new_vm) { 'newvm'}
+    let(:pool_name) { pool_object['name'] }
 
     before do
       expect(subject).not_to be_nil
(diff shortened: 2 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -577,21 +610,27 @@ def _check_pool(pool, provider)
...
       $redis.smembers("vmpooler__ready__#{pool['name']}").each do |vm|
         if inventory[vm]
           begin
+            pool_check_response[:checked_ready_vms] += 1
             check_ready_vm(vm, pool['name'], pool['ready_ttl'] || 0, provider)
           rescue => err
             $logger.log('d', "[!] [#{pool['name']}] _check_pool failed with an error while evaluating ready VMs: #{err}")
           end
+        else
+          move_vm_queue(pool['name'], vm, 'ready', 'completed', 'is a ready VM but is missing from inventory.  Marking as completed.')
```
sbeaulie: We could check when the VM started running and if it was more than a couple days ago, and it is also not in the inventory, then mark as completed.

glennsarti: so have an uber-timeout?

sschneid:
Might be easier to just set an `EXPIRE` on the redis keys?

glennsarti:
Yes and no.  Using expire would still mean the keys exist in REDIS if even we know that the VM doesn't exist.  Which means it can still hand out phantom VMs

glennsarti:
My reasoning: The provider inventory is "the source of truth" and the redis version of it, is just an information cache.  Therefore state should flow from the inventory into redis ASAP.

lib/vmpooler/pool_manager.rb
sbeaulie: thats how I would manually clean it... I would probably make sure it's an outdated 'running' VM.

glennsarti: How do you make sure?  Is this something the VM provider should surface or is the inventory process (`vms_in_pool`) enough?

glennsarti:
Looks like it needs a rebase

joshcooper:
Can we fix the underlying issues that cause vmpooler to get confused first? Otherwise, we trying to make decisions given incomplete data. For example, in the inventory loop, if we fail to connect to the vsphere `provider`, [we rescue the exception, log a warning, but we continue processing the loop!](https://github.com/glennsarti/vmpooler/blob/a555c06acaef54471a53c27cb00b135a2dc5fc81/lib/vmpooler/pool_manager.rb#L604-L621). So then a VM might not be in the inventory because we failed to connect or because it really isn't in vsphere anymore, but the pooler doesn't know which is the case.

glennsarti:
Yeah... at the very least, it should re raise and not check the pool at all.

glennsarti:
Rebased and updated for @joshcooper comment.  Added tests for that scenario.

lib/vmpooler/pool_manager.rb
```diff
@@ -603,6 +603,7 @@ def _check_pool(pool, provider)
         end
       rescue => err
         $logger.log('s', "[!] [#{pool['name']}] _check_pool failed with an error while inspecting inventory: #{err}")
+        return pool_check_response
```
sbeaulie: when exiting the loop, will the thread to loop for that pool be re-created?

glennsarti: As I am swallowing the error in this case, the loop thread will not terminate so it will not need to be recreated

Issue #230, opened 2017-07-18
Do not have a hardcoded list of VM providers


glennsarti:
The list of available VM providers is a hardcoded list in the `create_provider_object` method.  This should be changed so it load providers both from within the pooler itself, or available from gems.  This will make it easier to separate VM providers from the Pool Manager logic, and allow more varied providers.

glennsarti:
Fixed in https://github.com/puppetlabs/vmpooler/pull/231

PR #231, opened 2017-07-18
(GH-230) Dynamically load VM Providers

glennsarti:
Previously, a static list was used to instantiate VM Pooler Provider objects.
This commit changes the loader to instead interrogate the available clases in
the Vmpooler::PoolManager::Provider namespace and then instantiate from there.

This means class names are not case sensitive and that VM Providers can now be
dynamically loaded from other sources such as gems in the LOADPATH.  No tests
were added as this behaviour is exercised in the execute! tests already.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -735,21 +735,20 @@ module Vmpooler
       raise
     end
 
-    # create a provider object based on the providers/*.rb class that implements providers/base.rb
-    # provider_class: needs to match a provider class in providers/*.rb ie Vmpooler::PoolManager::Provider::X
-    # provider_name: should be a unique provider name
+    # Create a provider object, usually based on the providers/*.rb class, that implements providers/base.rb
+    # provider_class: Needs to match a class in the Vmpooler::PoolManager::Provider namespace. This is
+    #                 either as a gem in the LOADPATH or in providers/*.rb ie Vmpooler::PoolManager::Provider::X
+    # provider_name:  Should be a unique provider name
     #
     # returns an object Vmpooler::PoolManager::Provider::*
     # or raises an error if the class does not exist
     def create_provider_object(config, logger, metrics, provider_class, provider_name, options)
-      case provider_class
-      when 'vsphere'
-        Vmpooler::PoolManager::Provider::VSphere.new(config, logger, metrics, provider_name, options)
-      when 'dummy'
-        Vmpooler::PoolManager::Provider::Dummy.new(config, logger, metrics, provider_name, options)
-      else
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -735,21 +735,26 @@ def _check_pool(pool, provider)
...
-      when 'vsphere'
-        Vmpooler::PoolManager::Provider::VSphere.new(config, logger, metrics, provider_name, options)
-      when 'dummy'
-        Vmpooler::PoolManager::Provider::Dummy.new(config, logger, metrics, provider_name, options)
-      else
-        raise("Provider '#{provider_class}' is unknown for pool with provider '#{provider_name}'")
-      end
+      # Can't use just the string name due to old jruby support.  Instead, need
+      # to use the long form
+      provider_klass = Object.const_get('Vmpooler').const_get('PoolManager').const_get('Provider')
```
nwolfe: Since we know what class/module we're after here, do we really need the `const_get` for this part?

Or can this just be:
```ruby
provider_klass = Vmpooler::PoolManager::Provider
```

lib/vmpooler/pool_manager.rb
```diff
@@ -735,21 +735,26 @@ def _check_pool(pool, provider)
...
+      provider = nil
+      provider_klass.constants.each do |classname|
```
nwolfe: Why do we need to iterate over all the constants?

Can't we just basically `const_get` directly like we're doing on line 752?

glennsarti: Unfortunately const_get is case sensitive

nwolfe: So adjust the string before calling `const_get`, right?

glennsarti: if it's truly dynamic, the config file has no idea what the class name is e.g.  is it 'vsphere', 'VSPHERE', 'vSphere'.

The number of constants will probably be tiny (< 20) so there's no real speed difference here.

nwolfe: Oh `casecmp` is case-INSENSITIVE.....yeah that doesn't sound backwards

lib/vmpooler/pool_manager.rb
glennsarti: Might be right there...testing...

glennsarti: Yep 100% right.  Fixed and managed to shrink the method even smaller

glennsarti:
Yeah, this is mainly to allow community to write their own VM Providers as gems instead of having to integrate into this repo.

PR #232, opened 2017-07-18
Fix rubocop

glennsarti:
Bunch of rubocop fixes.  Only files left are in the api, but we don't have enough tests in that area to bravely fix them.


lib/vmpooler/pool_manager.rb
```diff
@@ -462,7 +460,7 @@ def _check_snapshot_queue
 
     def migration_limit(migration_limit)
       # Returns migration_limit setting when enabled
-      return false if migration_limit == 0 || ! migration_limit
+      return false if migration_limit == 0 || !migration_limit # rubocop:disable Style/NumericPredicate
```
glennsarti: Disable the cop here because migration limit can come in as an Integer OR boolean type

PR #236, opened 2017-09-01
(GH-226) Respond quickly to VMs being consumed

glennsarti:
Previously in commit 9b0e55f the looping period was changed from a static
number to a dynamic one depending on load, however this meant that the operation
to refill a pool was slowed down somewhat.  While not a problem under normal
loads, when a pool was quickly consumed, the pool manager may not respond
quickly enough to refill the pool.  This commit:

- Changes the sleep method, to us a helper sleep method that will wakeup
  periodically and evaluate other wakeup events.  This could be used later to
  exit sleep loops when pooler is shutting down to stop blocking threads
- By default the wakeup_period is set to the minimum pool check loop time, thus
  emulating the behaviour prior to commit 9b0e55f
- Adds tests for the behaviour

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,14 +12,33 @@ gem 'puma', '>= 3.6.0'
 gem 'rack', '~> 1.6'
 gem 'rake', '>= 10.4'
 gem 'rbvmomi', '>= 1.8'
-if RUBY_VERSION =~ /^1\.9\./
-  gem 'nokogiri', '< 1.7.0'
-end
-gem 'redis', '>= 3.2'
 gem 'sinatra', '>= 1.4'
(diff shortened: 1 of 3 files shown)


shrug:
@mattkirby ping, when you get a sec

shrug:
```
Gem::InstallError: redis requires Ruby version >= 2.2.2.
```
Do we need to pin the version of the redis gem?

glennsarti:
Doh...looks like we do :-(

I'll add this to the this PR

glennsarti:
@shrug Done.  Nokogiri needed some work too.

glennsarti:
Thanks @mattkirby

PR #239, opened 2017-09-28
Fix no implicit conversion to rational from nil

sbeaulie:
Before this change if the boottime was nil, the check_ready
loop would exit on Time.now - host['boottime'] with a TypeError
in jruby. The boottime is nil when the power is Off so moving that check
earlier should catch that bug.


underscorgan:
Is this ready for merge?

PR #241, opened 2017-10-09
Allow user to specify a configuration file in VMPOOLER_CONFIG_FILE variable

### LINKED ISSUE #240: YAML.safe_load does not work with symbols in config file
I have no idea how this ever worked.  YAML.safe_load does not like the usage of symbols in the config file.

```ruby
if ENV['VMPOOLER_CONFIG']
     # Load configuration from ENV
      parsed_config = YAML.safe_load(ENV['VMPOOLER_CONFIG'])
end
```

This throws error `Psych::DisallowedClass:  Tried to load unspecified class: Symbol`

The fix for now is to not use YAML.safe_load and instead just use YAML.load.

adamdav:
Previously, there were two ways to configure Vmpooler, either by
changing the contents of vmpooler.yaml or by assigning the raw YAML
to the VMPOOLER_CONFIG environment variable. This commit adds a new
environment variable called VMPOOLER_CONFIG_FILE that can be assigned
the name of a config file to use. Also fixes #240 by whitelisting the
Symbol class when calling YAML.safe_load in Vmpooler.config.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,6 @@
 Gemfile.lock
 Gemfile.local
 vendor
-vmpooler.yaml
+/vmpooler.yaml
 .bundle
 coverage
--- a/docs/dev-setup.md
+++ b/docs/dev-setup.md
@@ -62,17 +62,25 @@ C:\vmpooler > $ENV:VMPOOLER_DEBUG = 'true'
 
 ### `VMPOOLER_CONFIG`
 
-When `VMPOOLER_CONFIG` is set, vmpooler will read its configuration from the content of the environment variable instead of from the `vmpooler.yaml` configuration file.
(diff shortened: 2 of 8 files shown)


logicminds:
@adamdav There is a failure in the test because the vmpooler.yaml cannot be found.  You might want to mock that file in that test.

```
 1) Vmpooler .config when no config is given defaults to vmpooler.yaml
     Failure/Error: default_config = YAML.load_file(default_config_file)
     
     Errno::ENOENT:
       No such file or directory @ rb_sysopen - /home/travis/build/puppetlabs/vmpooler/spec/fixtures/vmpooler.yaml
     # ./spec/unit/vmpooler_spec.rb:17:in `block (4 levels) in <top (required)>'
```

adamdav:
@glennsarti branch is rebased

PR #242, opened 2017-10-14
(QENG-5305) Improve vmpooler host selection

mattkirby:
This change updates how vmpooler creates VMs, migrates VMs, and significantly reduces strain placed upon vsphere appliances by eliminating duplicate host state inspection.

1. Clone operations target the resource pool for a cluster instead of a specific host. Host selection is left for vsphere to determine.
2. Host selection used for migrating VMs is updated to eliminate duplicate host inspection by maintaining a global list of available target hosts within a cluster and updating it when it becomes stale.
3. VMs are evaluated for migration against the list instead of occupying a slot for migration to be evaluated.
4. A optional capability to create pool folders is added (this can be moved to a separate PR if desired)
5. Host object retrieval is updated to use the searchIndex instead of iterating through a list of cluster objects until the matching name is found.
6. A safeguard against no quickstats being found is added by ensuring that CPU utilization is not 0.

TODO:
- [x] Finish adding tests for new functionality in vsphere.rb


lib/vmpooler/pool_manager.rb
```diff
@@ -476,29 +550,32 @@ def migrate_vm(vm_name, pool_name, provider)
...
+        run_select_hosts(provider, pool_name, provider_name, vm['cluster'], vm['datacenter'], max_age, percentage_of_hosts_below_average)
+        if migration_count >= migration_limit
+          $logger.log('s', "[ ] [#{pool_name}] '#{vm_name}' is running on #{vm['host']}. No migration will be evaluated since the migration_limit has been reached")
+        elsif $provider_hosts[provider_name][vm['datacenter']][vm['cluster']]['architectures'][vm['architecture']].include?(vm['host'])
+          $logger.log('s', "[ ] [#{pool_name}] No migration required for '#{vm_name}' running on #{vm['host']}")
         else
-          finish = migrate_vm_and_record_timing(vm_name, pool_name, parent_host_name, host_name, provider)
-          $logger.log('s', "[>] [#{pool_name}] '#{vm_name}' migrated from #{parent_host_name} to #{host_name} in #{finish} seconds")
+          $redis.sadd('vmpooler__migration', vm_name)
+          target_host_name = select_next_host(provider_name, vm['datacenter'], vm['cluster'], vm['architecture'])
```
sbeaulie: what if `select_next_host` returns 'nil'?

lib/vmpooler/pool_manager.rb
```diff
@@ -458,6 +461,77 @@ def _check_snapshot_queue
...
+        unless maxloop.zero?
+          break if loop_count >= maxloop
+          loop_count += 1
+        end
+      end
+    end
+
+    def select_next_host(provider_name, datacenter, cluster, architecture)
+      provider_hosts = $provider_hosts
+      host = provider_hosts[provider_name][datacenter][cluster]['architectures'][architecture][0]
```
sbeaulie: I think if `provider_hosts[provider_name][datacenter][cluster]['architectures'][architecture]` is a string then `[0]` could return the first character which is not what you probably expect here?

lib/vmpooler/providers/vsphere.rb
```diff
@@ -579,6 +585,65 @@ def memory_utilization_for(host)
...
+        def build_compatible_hosts_lists(hosts, percentage)
+          hosts_with_arch_versions = hosts.map { |host| [host[0], host[1], get_host_cpu_arch_version(host[1])] }
+          versions = hosts_with_arch_versions.map { |host| host[2] }.uniq
+          architectures = {}
+          versions.each do |version|
+            architectures[version] = []
+          end
+
+          hosts_with_arch_versions.each do |host|
+            architectures[host[2]] << [host[0], host[1], host[2]]
```
sbeaulie: I find it hard to know what is in each index `host[0], host[1], host[2]`

lib/vmpooler/pool_manager.rb
mattkirby: It should cause a failure to be logged reporting that vm migration failed with an error raised from lib/vmpooler/providers/vsphere.rb:90 when migrate_vm is called. Perhaps it's worth emitting a more clear error that there was no target host provided.

lib/vmpooler/providers/vsphere.rb
mattkirby: Yeah, it probably makes sense to put it in a hash and give these values some titles.

lib/vmpooler/pool_manager.rb
mattkirby: That's true. It's populated by select_least_used_hosts from lib/vmpooler/providers/vsphere.rb:607 which returns an array. Do you think I should be more explicit than this at some point to ensure an array is returned?

sbeaulie: Ok I got it, thanks

lib/vmpooler/pool_manager.rb
sbeaulie: I think it will call the next line `migrate_vm_and_record_timing()` with `target_host_name` = nil which seems to go all the way  down to the vcenter API to try and get the host by dns name? 
We should fail earlier than that.

lib/vmpooler/pool_manager.rb
```diff
@@ -21,6 +21,9 @@ def initialize(config, logger, redis, metrics)
 
       # Our thread-tracker object
       $threads = {}
+
+      # Host tracking object
+      $provider_hosts = {}
```
glennsarti: nit: REALLY REALLY try to avoid adding more global variables.  Instance level should be enough.

mattkirby: @glennsarti are you sure? I specifically require that this variable be accessible across all vmpooler provider threads.

mattkirby:
@glennsarti how do you think this can be solved in the vsphere code when it needs to constrain and share data across pool manager operations?

glennsarti:
Sure
* Only one instance of a provider exists for one or more pools so it's able to maintain the state inside the provider
* All operations have to go through the provider so they can be gated through there

The only reason I can think of to keep state in the pool manager is to allow across-provider communication.  Which is not (should not) be required for vsphere host selection.

mattkirby:
Ok. I have reworked this to do away with global variables and now use an instance variable in the provider to track host state on a per-provider basis. Tests pass, but I need to complete adding coverage for changes in vsphere and the base provider, but wanted to get this put here so I could get further feedback and input if there are still some clear changes needed. I still intend to address the other feedback.

mattkirby:
I've pushed an update here that moves migrate_vm to the vsphere provider. This should be a lot cleaner. I've addressed the confusing array population that @sbeaulie brought up earlier, and global variables are no longer used here in favor of instance variables on the provider, which works great. A mutex is added to ensure that all writing to provider_hosts in the provider is safe across all threads.

I'm still working through a handful of remaining tests on the vsphere provider, but this is all functional now.

lib/vmpooler/pool_manager.rb
mattkirby: This will now fail if the datacenter, or cluster are missing, and if the target is empty, before trying to run the methods with bad data.

lib/vmpooler/providers/vsphere.rb
mattkirby: This now uses a hash which should make clear what's being placed into this array.

mattkirby:
The jruby tests pass, the others are failing. If I'm reading travis correctly it looks as though setup to run the tests is what's failing.

lib/vmpooler/pool_manager.rb
```diff
@@ -138,6 +129,14 @@ def _check_ready_vm(vm, pool, ttl, provider)
         end
       end
 
+      $redis.hset('vmpooler__vm__' + vm, 'check', Time.now)
```
mattkirby: In my mess of rebasing and squashing I've managed to make it look like this change is part of my change, but this was introduced in #239

lib/vmpooler/pool_manager.rb
```diff
@@ -459,69 +458,17 @@ def _check_snapshot_queue
...
 
-    def migration_limit(migration_limit)
-      # Returns migration_limit setting when enabled
-      return false if migration_limit == 0 || !migration_limit # rubocop:disable Style/NumericPredicate
-      migration_limit if migration_limit >= 1
-    end
-
     def migrate_vm(vm_name, pool_name, provider)
       Thread.new do
         begin
```
glennsarti: The provider really shouldn't need to contact redis for this.  In this case the provider is just removing it from the queue.  So the `$redis.srem('vmpooler__migrating__' + pool_name, vm_name)` should be called here PRIOR to the `provider.migrate_vm` call

lib/vmpooler/pool_manager.rb
```diff
@@ -459,69 +458,17 @@ def _check_snapshot_queue
...
-          _migrate_vm(vm_name, pool_name, provider)
+          provider.migrate_vm(pool_name, vm_name, $redis)
         rescue => err
           $logger.log('s', "[x] [#{pool_name}] '#{vm_name}' migration failed with an error: #{err}")
-          remove_vmpooler_migration_vm(pool_name, vm_name)
+          provider.remove_vmpooler_migration_vm(pool_name, vm_name, $redis)
```
glennsarti: This new method `remove_vmpooler_migration_vm` doesn't exist on the base and on inspection probably shouldn't be in a provider anyway.  It's just removing the VM from the task queue.

By doing my previous suggestion, it's not needed as the VM is removed from the queue prior to be actioned in the provider.

lib/vmpooler/pool_manager.rb
```diff
@@ -459,69 +458,17 @@ def _check_snapshot_queue
       end
     end
 
-    def migration_limit(migration_limit)
```
glennsarti: As the migration limit is now a per vSphere provider setting the documentation should be updated too.

lib/vmpooler/providers/base.rb
```diff
@@ -11,12 +11,16 @@ class Base
         attr_reader :metrics
         # Provider options passed in during initialization
         attr_reader :provider_options
+        # Hash for tracking hosts for deployment
```
glennsarti: These settings seem vsphere specific.  Perhaps instead of modifying base, which affects all providers, move them there

lib/vmpooler/providers/base.rb
```diff
@@ -11,12 +11,16 @@ class Base
...
+        attr_reader :provider_hosts
 
         def initialize(config, logger, metrics, name, options)
           @config = config
           @logger = logger
           @metrics = metrics
           @provider_name = name
+          @provider_hosts = {}
```
glennsarti: As above.  vsphere provider specific.

lib/vmpooler/providers/base.rb
```diff
@@ -120,6 +124,14 @@ def migrate_vm_to_host(_pool_name, _vm_name, _dest_host_name)
         end
 
         # inputs
+        #   [String] pool_name      : Name of the pool
+        #   [String] vm_name        : Name of the VM to migrate
+        #   [Class] redis           : Redis object
+        def migrate_vm(_pool_name, _vm_name, _redis)
```
glennsarti: Apart from Redis, how is this method signature different from `migrate_vm_to_host`?.  Is it just simply calling

`migrate_vm_to_host(pool_name, vm_name, nil)` ?

Redis is already a global variable, so no ed to pass it through.  That should probably be a separate PR

lib/vmpooler/pool_manager.rb
```diff
@@ -459,69 +458,17 @@ def _check_snapshot_queue
...
-      # Returns migration_limit setting when enabled
-      return false if migration_limit == 0 || !migration_limit # rubocop:disable Style/NumericPredicate
-      migration_limit if migration_limit >= 1
-    end
-
     def migrate_vm(vm_name, pool_name, provider)
```
glennsarti: There is currently no way in Pool Manager to actually add a VM to the migration queue.  So this PR will not have any affect yet.  Is this by design?

lib/vmpooler/providers/base.rb
mattkirby: The main difference to me, beyond the redis part which as you point out shouldn't be there, is that there's no longer a need to provide the destination host name, so I figured keeping migrate_vm instead of migrate_vm_to_host made sense. I can change it if it makes more sense.

lib/vmpooler/pool_manager.rb
mattkirby: No, it's an oversight. I'll fix it.

lib/vmpooler/pool_manager.rb
mattkirby: It probably should be a per vSphere provider setting, but it's still tracking migrations in redis and getting the setting from the global configuration section instead of checking the provider for this setting. I could add this change here. I was considering tackling that and the task_limit concept in a future change.

glennsarti: Could be as simple as add in the docs `migration limit (Only affects vSphere Provider)`

mattkirby: Oh I see what you're saying, not that it's a per provider setting, but rather that it is only a vsphere provider setting.

glennsarti: yeah...  The docs suggest its global but no longer.

mattkirby: I added a mention in the example configuration file. I couldn't find another reference to migration_limit, please let me know if I've missed another location this should go.

lib/vmpooler/pool_manager.rb
mattkirby: I took your suggestion and placed this prior to calling `provider.migrate_vm`.

mattkirby:
I've updated this to include a fix for a behavior witnessed when running this over the weekend. Specifically, if no hosts are available for deployment check_time_finished is still updated in order to allow for things waiting on host selection to proceed. Previously, they would be stuck waiting and gum up pool_manager. Now, they will fail fast and report an error that makes clear why the clone or migration failed. Additionally, settings for utilization maximum and max age are exposed and documented.

shrug:
@mattkirby Ready to merge this?

mattkirby:
Just about. I've got one more set of tests I'm going to push shortly after which this should be ready to go. That should cover the bits I've added that haven't had tests added yet.

mattkirby:
I've finished tests and this should be ready to go.

PR #244, opened 2017-10-18
(POOLER-92) Add the alias information in the API status page for each…

sbeaulie:
… pool

Before this change if a pool had an alias configured, the information would not be
made public in the API. This commit adds the alias key in the pool object for each
pool if configured. The alias key can be absent, a string or an one or multiple
array of strings. The value of the alias is copied from the configuration and can
represent another name for the pool, or another configured pool.


vmpooler.yaml.aliasedpools
```diff
@@ -0,0 +1,74 @@
+---
```
glennsarti: Is this test fixture file in the right location? i.e. the root of the project?  Should it be under `/spec/`?

sbeaulie: It's not used by the spec tests, in fact I have not seen any `.yaml` file in the spec directory. I want that example dummy config file in the repo to share that file with my team members working on integrating this new information into the ABS scheduler. I think the directory `ex` would be better for this?

glennsarti: Perhaps the `docs` dir?

PR #246, opened 2017-11-28
(POOLER-96) Setting the Rubygems version

sbeaulie:
travis uses the latest bundler 1.16 which installs the newer rubygems and I think that might be why it's failing now when downgrading only bundler to version 1.12
This is a test to see if travis will pass.


sbeaulie:
Travis now passes, feel free to squash merge

sbeaulie:
this reflects our production better as we use bundler 1.15.4 and jruby 1.7.13

PR #248, opened 2018-01-08
Identify when ESXi host quickstats do not return

mattkirby:
This commit updates cpu_utilization_for and memory_utilization_for to detect when quickstats are not present. Without this change a nil result is transformed to 0, which is perceived as a host that has no utilization.

### DIFF
--- a/lib/vmpooler/providers/vsphere.rb
+++ b/lib/vmpooler/providers/vsphere.rb
@@ -625,8 +625,11 @@ module Vmpooler
           cpu_utilization = cpu_utilization_for host
           memory_utilization = memory_utilization_for host
 
-          return nil if cpu_utilization == 0
-          return nil if memory_utilization == 0
+          return nil if cpu_utilization.nil?
+          return nil if cpu_utilization == 0.0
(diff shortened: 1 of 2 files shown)


lib/vmpooler/providers/vsphere.rb
```diff
@@ -569,12 +574,14 @@ def get_host_cpu_arch_version(host)
...
         def cpu_utilization_for(host)
           cpu_usage = host.summary.quickStats.overallCpuUsage
+          return nil if cpu_usage.nil?
           cpu_size = host.summary.hardware.cpuMhz * host.summary.hardware.numCpuCores
           (cpu_usage.to_f / cpu_size.to_f) * 100
         end
 
         def memory_utilization_for(host)
           memory_usage = host.summary.quickStats.overallMemoryUsage
+          rturn nil if memory_usage.nil?
```
demophoon: `return`

sbeaulie:
I'm good with this change once the typo is fixed

mattkirby:
Fixed the typo.

PR #250, opened 2018-03-23
(maint) Add the last boot time for each pool

sbeaulie:
This commit add a redis hash where there is one key per pool, and the
stored value is the last time a VM was booted e.g. the last time
a VM went from 'pending' to 'ready'. This is also displayed in the
API as lastBoot:'2018-03-23 17:43:39 +0000'. The data can then be
used by any external system, in this case our alarming system.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -205,17 +205,19 @@ module Vmpooler
       result[:pools] = {} unless views and not views.include?("pools")
       pools.each do |pool|
         # REMIND: move this out of the API and into the back-end
-        ready   = backend.scard('vmpooler__ready__' + pool['name']).to_i
-        running = backend.scard('vmpooler__running__' + pool['name']).to_i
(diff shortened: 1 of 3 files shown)


sbeaulie:
The redis data looks like
```
root@11ba491563b2:/var/lib/vmpooler# redis-cli
127.0.0.1:6379> keys vmpooler__lastboot*
1) "vmpooler__lastboot"
127.0.0.1:6379> hgetall vmpooler__lastboot
1) "debian-7-i386"
2) "2018-03-23 17:15:15 +0000"
3) "debian-7-x86_64"
4) "2018-03-23 17:15:15 +0000"
```
API looks like
screen shot 2018-03-23 at 12 43 59

underscorgan:
@sbeaulie this looks reasonable, but travis is angry

sbeaulie:
I reran the one ruby 2.1.1 cell that was failing and it passed 🤷‍♂️

PR #251, opened 2018-04-17
(POOLER-68) Replace find_vm search mechanism

mattkirby:
This commit replaces find_vm and find_vm_heavy with a more performant and reliable mechanism of identifying VM objects. Specifically, FindByInventoryPath is able to leverage known data about a VM, its folder path and datacenter, and use that to identify whether that VM exists by its location. Without this change find_vm_heavy is called each time a VM cannot be found, which is frequent, and in doing so uses PropertyCollector in a manner that is not thread-safe. Additionally, this PropertyCollector usage does not clean up its traces, which can cause vCenter appliance instability issues on VCSA 6.x.

### DIFF
--- a/lib/vmpooler/providers/vsphere.rb
+++ b/lib/vmpooler/providers/vsphere.rb
@@ -165,7 +165,7 @@ module Vmpooler
           vm_hash = nil
           @connection_pool.with_metrics do |pool_object|
             connection = ensured_vsphere_connection(pool_object)
-            vm_object = find_vm(vm_name, connection)
+            vm_object = find_vm(_pool_name, vm_name, connection)
             return vm_hash if vm_object.nil?
 
             vm_folder_path = get_vm_folder_path(vm_object)
@@ -283,7 +283,7 @@ module Vmpooler
 
           @connection_pool.with_metrics do |pool_object|
             connection = ensured_vsphere_connection(pool_object)
-            vm_object = find_vm(vm_name, connection)
+            vm_object = find_vm(pool_name, vm_name, connection)
             raise("VM #{vm_name} in pool #{pool_name} does not exist for the provider #{name}") if vm_object.nil?
 
             add_disk(vm_object, disk_size, datastore_name, connection, get_target_datacenter_from_config(pool_name))
@@ -294,7 +294,7 @@ module Vmpooler
         def create_snapshot(pool_name, vm_name, new_snapshot_name)
           @connection_pool.with_metrics do |pool_object|
             connection = ensured_vsphere_connection(pool_object)
-            vm_object = find_vm(vm_name, connection)
+            vm_object = find_vm(pool_name, vm_name, connection)
             raise("VM #{vm_name} in pool #{pool_name} does not exist for the provider #{name}") if vm_object.nil?
 
             old_snap = find_snapshot(vm_object, new_snapshot_name)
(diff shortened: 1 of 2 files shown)


lib/vmpooler/providers/vsphere.rb
```diff
@@ -784,60 +784,24 @@ def find_snapshot(vm, snapshotname)
...
-          while results.token
-            results = property_collector.ContinueRetrievePropertiesEx(token: results.token)
-            results.objects.each do |result|
-              name = result.propSet.first.val
-              next unless vmname.include? name
-              vms[name] = result.obj
-            end
-          end
+        def find_vm(pool_name, vmname, connection)
+          pool_configuration = pool_config(pool_name)
```
sbeaulie: this method can return nil so we should protect against that and maybe log an error message if it happens

lib/vmpooler/providers/vsphere.rb
```diff
@@ -784,60 +784,24 @@ def find_snapshot(vm, snapshotname)
...
+          folder = pool_configuration['folder']
+          datacenter = get_target_datacenter_from_config(pool_name)
```
sbeaulie: same comment here the method can return nil so we should protect against that and return an error message if it happens.

Additionally, we could have `find_vm` return nil, similarily to when `connection.searchIndex.FindByInventoryPath(propSpecs)` reutrns nil, but I haven't checked how we handle that upstream

lib/vmpooler/providers/vsphere.rb
```diff
@@ -165,7 +165,7 @@ def get_vm(_pool_name, vm_name)
           vm_hash = nil
           @connection_pool.with_metrics do |pool_object|
             connection = ensured_vsphere_connection(pool_object)
-            vm_object = find_vm(vm_name, connection)
+            vm_object = find_vm(_pool_name, vm_name, connection)
```
glennsarti: The ruby convention of underscore prefix is denote variables which aren't used.  As this variable is not being used, the underscore should be removed.

lib/vmpooler/providers/vsphere.rb
```diff
@@ -327,7 +327,7 @@ def revert_snapshot(pool_name, vm_name, snapshot_name)
         def destroy_vm(_pool_name, vm_name)
           @connection_pool.with_metrics do |pool_object|
             connection = ensured_vsphere_connection(pool_object)
-            vm_object = find_vm(vm_name, connection)
+            vm_object = find_vm(_pool_name, vm_name, connection)
```
glennsarti: See underscore comment above

lib/vmpooler/providers/vsphere.rb
glennsarti: I think it's more about pool_configuration being `nil` and the next line `pool_configuration['folder']` will raise a method error.

* Should either raise an error it pool_configuration is nil
`raise("could not find pool_name") if pool_configuration.nil?`

or

* Return a nil value
`return nil if pool_configuration.nil?`

glennsarti: Should probably document the expected return values in code comments for future-us

glennsarti: Actually given;

`expect(subject.find_vm(poolname,missing_vm,connection)).to be_nil` in the spec tests, it should return `nil`

lib/vmpooler/providers/vsphere.rb
mattkirby: Thanks, I'll fix this and push a change.

lib/vmpooler/providers/vsphere.rb
mattkirby: Ah ok, I follow now. I'll update this and push a change.

mattkirby:
Thanks for the review @glennsarti and @sbeaulie I think I've made appropriate adjustments here. Let me know if you think this requires further changes.

PR #256, opened 2018-05-22
(POOLER-107) Add configuration API endpoint

mattkirby:
This commit adds a configuration endpoint to the vmpooler API. Pool
size, and pool template, can be adjusted for pools that are configured
at vmpooler application start time. Pool template changes trigger a pool
refresh, and the new template has delta disks created automatically by
vmpooler.

Additionally, the capability to create template delta disks is added to the vsphere provider, and this is implemented to ensure that templates have delta disks created at application start time.

The mechanism used to find template VM objects is simplified to make the flow of logic easier to understand. As an additional benefit, performance of this lookup is improved by using FindByInventoryPath.

Lastly, API.md is updated to add a table of contents.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,8 +1,17 @@
-### API
+# Table of contents
+1. [API](#API)
+2. [Token operations](#token)
+3. [VM operations](#vmops)
+4. [Add disks](#adddisks)
+5. [VM snapshots](#vmsnapshots)
+6. [Status and metrics](#statusmetrics)
+7. [Pool configuration](#poolconfig)
+
+### API <a name="API"></a>
 
 vmpooler provides a REST API for VM management.  The following examples use `curl` for communication.
 
-#### Token operations
+#### Token operations <a name="token"></a>
 
 Token-based authentication can be used when requesting or modifying VMs.  The `/token` route can be used to create, query, or delete tokens.  See the provided YAML configuration example, [vmpooler.yaml.example](vmpooler.yaml.example), for information on configuring an authentication store to use when performing token operations.
 
@@ -76,7 +85,7 @@ Enter host password for user 'jdoe':
 }
 ```
 
-#### VM operations
+#### VM operations <a name="vmops"></a>
 
 ##### GET /vm
 
@@ -230,7 +239,7 @@ $ curl -X DELETE --url vmpooler.company.com/api/v1/vm/fq6qlpjlsskycq6
 }
 ```
 
-#### Adding additional disk(s)
+#### Adding additional disk(s) <a name="adddisks"></a>
 
 ##### POST /vm/&lt;hostname&gt;/disk/&lt;size&gt;
 
@@ -270,7 +279,7 @@ $ curl --url vmpooler.company.com/api/v1/vm/fq6qlpjlsskycq6
 
 ````
 
-#### VM snapshots
+#### VM snapshots <a name="vmsnapshots"></a>
 
 ##### POST /vm/&lt;hostname&gt;/snapshot
 
@@ -322,7 +331,7 @@ $ curl X POST -H X-AUTH-TOKEN:a9znth9dn01t416hrguu56ze37t790bl --url vmpooler.co
 }
 ````
 
-#### Status and metrics
+#### Status and metrics <a name="statusmetrics"></a>
 
 ##### GET /status
 
@@ -538,5 +547,99 @@ $ curl -G -d 'from=2015-03-10' -d 'to=2015-03-11' --url vmpooler.company.com/api
       "tag": { }
     }
   ]
+}
+```
+
+#### Managing pool configuration via API <a name="poolconfig"></a>
+
+##### GET /config
+
+Returns the running pool configuration
+
+Responses:
+* 200 - OK
+* 404 - No configuration found
+```
+$ curl https://vmpooler.company.com/api/v1/config
+```
+```json
+{
+  "pool_configuration": [
+    {
+      "name": "redhat-7-x86_64",
+      "template": "templates/redhat-7.2-x86_64-0.0.3",
+      "folder": "vmpooler/redhat-7-x86_64",
+      "datastore": "stor1",
+      "size": 1,
+      "datacenter": "dc1",
+      "provider": "vsphere",
+      "capacity": 1,
+      "major": "redhat",
+      "template_ready": true
+    }
+  ],
+  "status": {
+    "ok": true
+  }
+}
+```
+
+Note: to enable poolsize and pooltemplate config endpoints it is necessary to set 'experimental_features: true' in your vmpooler configuration. A 405 is returned when you attempt to interact with these endpoints when this configuration option is not set.
+
+##### POST /config/poolsize
+
+Change pool size without having to restart the service.
+
+All pool template changes requested must be for pools that exist in the vmpooler configuration running, or a 404 code will be returned
+
+When a pool size is changed due to the configuration posted a 201 status will be returned. When the pool configuration is valid, but will not result in any changes, 200 is returned.
+
+Pool size configuration changes persist through application restarts, and take precedence over a pool size value configured in the pool configuration provided when the application starts. This persistence is dependent on redis. So, if the redis data is lost then the configuration updates revert to those provided at startup at the next application start.
+
+An authentication token is required in order to change pool configuration when authentication is configured.
+Responses:
+* 200 - No changes required
+* 201 - Changes made on at least one pool with changes requested
+* 400 - An invalid configuration was provided causing requested changes to fail
+* 404 - An unknown error occurred
+* 405 - The endpoint is disabled because experimental features are disabled
+```
+$ curl -X POST -H "Content-Type: application/json" -d '{"debian-7-i386":"2","debian-7-x86_64":"1"}' --url https://vmpooler.company.com/api/v1/config/poolsize
+```
+```json
+{
+  "ok": true
+}
+```
+
+##### POST /config/pooltemplate
+
+Change the template configured for a pool, and replenish the pool with instances built from the new template.
+
+All pool template changes requested must be for pools that exist in the vmpooler configuration running, or a 404 code will be returned
+
+When a pool template is changed due to the configuration posted a 201 status will be returned. When the pool configuration is valid, but will not result in any changes, 200 is returned.
+
+A pool template being updated will cause the following actions, which are logged in vmpooler.log:
+* Destroy all instances for the pool template being updated that are in the ready and pending state
+* Halt repopulating the pool while creating template deltas for the newly configured template
+* Unblock pool population and let the pool replenish with instances based on the newly configured template
+
+Pool template changes persist through application restarts, and take precedence over a pool template configured in the pool configuration provided when the application starts. This persistence is dependent on redis. As a result, if the redis data is lost then the configuration values revert to those provided at startup at the next application start.
+
+An authentication token is required in order to change pool configuration when authentication is configured.
+
+Responses:
+* 200 - No changes required
+* 201 - Changes made on at least one pool with changes requested
+* 400 - An invalid configuration was provided causing requested changes to fail
+* 404 - An unknown error occurred
+* 405 - The endpoint is disabled because experimental features are disabled
+```
+$ curl -X POST -H "Content-Type: application/json" -d '{"debian-7-i386":"templates/debian-7-i386"}' --url https://vmpooler.company.com/api/v1/config/pooltemplate
+```
+```json
+{
+  "ok": true
 }
 ```
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -120,6 +120,74 @@ module Vmpooler
       result
     end
 
+    def update_pool_size(payload)
+      result = { 'ok' => false }
+
+      pool_index = pool_index(pools)
+      pools_updated = 0
+      sync_pool_sizes
+
+      payload.each do |poolname, size|
+        unless pools[pool_index[poolname]]['size'] == size.to_i
+          pools[pool_index[poolname]]['size'] = size.to_i
+          backend.hset('vmpooler__config__poolsize', poolname, size)
+          pools_updated += 1
+          status 201
+        end
+      end
+      status 200 unless pools_updated > 0
+      result['ok'] = true
+      result
+    end
+
+    def update_pool_template(payload)
+      result = { 'ok' => false }
+
+      pool_index = pool_index(pools)
+      pools_updated = 0
+      sync_pool_templates
+
+      payload.each do |poolname, template|
+        unless pools[pool_index[poolname]]['template'] == template
+          pools[pool_index[poolname]]['template'] = template
+          backend.hset('vmpooler__config__template', poolname, template)
+          pools_updated += 1
+          status 201
+        end
+      end
+      status 200 unless pools_updated > 0
+      result['ok'] = true
+      result
+    end
+
+    def sync_pool_templates
+      pool_index = pool_index(pools)
+      template_configs = backend.hgetall('vmpooler__config__template')
+      unless template_configs.nil?
+        template_configs.each do |poolname, template|
+          if pool_index.include? poolname
+            unless pools[pool_index[poolname]]['template'] == template
+              pools[pool_index[poolname]]['template'] = template
+            end
+          end
+        end
+      end
+    end
+
+    def sync_pool_sizes
+      pool_index = pool_index(pools)
+      poolsize_configs = backend.hgetall('vmpooler__config__poolsize')
+      unless poolsize_configs.nil?
+        poolsize_configs.each do |poolname, size|
+          if pool_index.include? poolname
+            unless pools[pool_index[poolname]]['size'] == size.to_i
+              pools[pool_index[poolname]]['size'] == size.to_i
+            end
+          end
+        end
+      end
+    end
+
     # Provide run-time statistics
     #
     # Example:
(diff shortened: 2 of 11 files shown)


glennsarti:
Does this need some RBAC? i.e. can anyone drain a pool?

mattkirby:
Yes, I think some sort of RBAC is warranted. I have some ideas for how to implement that, but this change was getting pretty large, so I wanted to put it up for review and consider handling that as a separate change. With that said, if that doesn't sound like a reasonable approach I'm open to blocking this change on implementing some sort of RBAC. I have some ideas, but have not started digging into the details of what would be required.

glennsarti:
You could also make those API endpoints opt in too (Experimental Feature).  So not enabled by default.

As for RBAC.  You could start simpler and just list which tokens have a "Global Admin" role in the config file.  Is there an issue/ticket for this as we're probably better off putting this info there.

lib/vmpooler/api/v1.rb
```diff
@@ -747,6 +803,58 @@ def invalid_templates(payload)
...
+      content_type :json
+      result = { 'ok' => false }
+
+      need_token! if Vmpooler::API.settings.config[:auth]
+
+      payload = JSON.parse(request.body.read)
+
+      if payload
+        invalid = invalid_template_or_size(payload)
+        if invalid.empty?
```
sbeaulie: this cancels the whole request when at least one pool or size is invalid. Not sure if that's what you intend to do, as the other option is the accept the ones that are valid.

lib/vmpooler/api/v1.rb
```diff
@@ -747,6 +803,58 @@ def invalid_templates(payload)
...
+          result = update_pool_size(payload)
+        else
+          invalid.each do |bad_template|
+            metrics.increment("config.invalid.#{bad_template}")
+          end
+          status 404
```
sbeaulie: I would add the content of the `invalid` array you built
something like
```
 result[:message] = invalid.map { |i| "Invalid pool size change: #{i}" }.join(",")
```

lib/vmpooler/api/v1.rb
mattkirby: Yes, that's what I intended to do since that is the pattern followed with the vm endpoint and it keeps things simple. I'm open to handling these cases in a more complex fashion, but wanted to minimize initial complexity.

lib/vmpooler/api/v1.rb
mchllweeks: +1 for adding a message to the result.  Would it also be helpful to add some logging info and error statements besides the metrics?  Just thinking about possibly needing to go back through the logs to validate whether or not a request was successful.

lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +555,75 @@ def check_pool(pool,
       end
     end
 
+    def prepare_template(pool, provider)
+      if $config[:config]['create_template_delta_disks']
+        # Ensure templates are evaluated for delta disk creation on startup
+        return if $redis.hget('vmpooler__config__updating', pool['name'])
+        unless $redis.hget('vmpooler__template__prepared', pool['name'])
```
sbeaulie: What hapens when I update the template say from -0.0.1 to -0.0.2? My understanding is this `unless` will skip because it's already been done for 0.0.1, which is not what we want

lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +555,75 @@ def check_pool(pool,
...
+          begin
+            $redis.hset('vmpooler__config__updating', pool['name'], 1)
+            provider.create_template_delta_disks(pool)
```
sbeaulie: this is open for multiple thread access (we have multiple threads per pool right?). My worry is that `vmpooler__config__updating` is not atomic and a race scenario is possible here. What's the worst case if a few threads run `create_template_delta_disks` in parallel?

If this is just to make sure we are not wasting resources it might be good enough to live with this exposure?

lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +555,75 @@ def check_pool(pool,
...
+          $redis.hset('vmpooler__config__updating', pool['name'], 1)
+
+          old_template_name = $redis.hget('vmpooler__template', pool['name'])
+          new_template_name = $redis.hget('vmpooler__config__template', pool['name'])
+          pool['template'] = new_template_name
+          $redis.hset('vmpooler__template', pool['name'], new_template_name)
+          $logger.log('s', "[*] [#{pool['name']}] template updated from #{old_template_name} to #{new_template_name}")
+          # Remove all ready and pending VMs so new instances are created from the new template
+          if $redis.scard("vmpooler__ready__#{pool['name']}") > 0
+            $logger.log('s', "[*] [#{pool['name']}] removing ready and pending instances")
```
sbeaulie: in practive this log message will be triggered for 'ready' only and not 'pending'

lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +555,75 @@ def check_pool(pool,
...
+            $redis.smembers("vmpooler__ready__#{pool['name']}").each do |vm|
+              $redis.smove("vmpooler__ready__#{pool['name']}", "vmpooler__completed__#{pool['name']}", vm)
```
sbeaulie: should we use the method `move_vm_queue()`

lib/vmpooler/pool_manager.rb
```diff
@@ -683,36 +752,59 @@ def _check_pool(pool, provider)
         end
       end
 
+      # Create template delta disks
+      prepare_template(pool, provider)
```
sbeaulie: if we just want to do this on startup, maybe we could move it out of this running loop and go over each pool on startup? that would remove multiple thread access.

lib/vmpooler/pool_manager.rb
```diff
@@ -683,36 +752,59 @@ def _check_pool(pool, provider)
...
+                $redis.decr('vmpooler__tasks__clone')
+                raise
+              end
             end
           end
         end
       end
 
+      # Remove VMs in excess of the configured pool size
+      remove_excess_vms(pool, provider, ready, total)
```
sbeaulie: The problem again is that we could hit this twice, and say each thread removes the extra 2 Vms over the total, we could be down 4 Vms. One way to reduce this exposure would be to add locks or to have the total and ready numbers checked for every single vm removal.

lib/vmpooler/pool_manager.rb
mattkirby: We could, though if it can happen in parallel I think startup time should be better. I had some brief internal debate about whether it worked better in check_pool, or not, and arrived at this being the solution that worked. I'm open to consider it further though if this doesn't make sense.

lib/vmpooler/pool_manager.rb
mattkirby: It happens fast enough that I don't think this ends up being a concern in real world usage. We can try to cause a failure here and if we can make one happen we could try to implement some safeguards. With that said I have found that I have not been able to cause this to make a mistake purposefully.

lib/vmpooler/api/v1.rb
mattkirby: We will get log messages from the pool_manager interactions. We don't otherwise track success or failures of API operations, or log anything from the API currently.

lib/vmpooler/pool_manager.rb
mattkirby: In this case `update_pool_template` will handle the pool update, where the pool will be flushed, it will once again evaluate the template for delta prepare disk creation, and then allow pool repopulation with the new template. It's true that `prepare_template` will not handle this case. It's perhaps a little confusing, but it is by design so they can handle their processes separately.

lib/vmpooler/pool_manager.rb
mattkirby: In my testing I have seen that these sorts of safeguards successfully prevent duplicate instances of a task from running. In this context a delta disk is only being created for the template configured for the specific pool running the task. If two instances of this run I don't think there will be any issue as we can run create_delta_disks repeatedly without any discernible issue.

lib/vmpooler/pool_manager.rb
mattkirby: Yes, I should probably change it. I'm not sure if it's helpful to log that pending and ready instances are being removed individually as well, or not.

lib/vmpooler/pool_manager.rb
mattkirby: We probably should. I've moved other smove use cases to use this method, so I may as well add this.

lib/vmpooler/pool_manager.rb
mattkirby: I have updated this to use a mutex object to synchronize these events so that only a single instance can happen across all threads. This should be more robust, and safe, than the redis based checking. I don't think that the methods used would have been likely to cause concern, but nonetheless this is more specific and should ensure the expected behavior.

docs/API.md
```diff
@@ -540,3 +549,59 @@ $ curl -G -d 'from=2015-03-10' -d 'to=2015-03-11' --url vmpooler.company.com/api
   ]
 }
 ```
+
+#### Changing configuration via API <a name="poolconfig"></a>
+
+##### POST /config/poolsize
```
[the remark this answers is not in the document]
mattkirby: We could consider alternately using a single endpoint. I started with poolsize and then added pooltemplate. Part of my goal here has been to minimize complexity so I thought having two separate paths under /config seemed reasonable for that. With that said, I'm open to change it if it doesn't make sense.

As of right now I hadn't made a plan to add pools that aren't configured. If it's something we want to support we could add it. My thought is that it allows the configuration file to serve as a source of truth about which pools are configured, even though size and template configurations may change.

I think it would be possible to add pools live, but haven't determined exactly what additional work would be required to do it. In the cases setup now there's already a check_pool thread running for each pool that exists to detect and deploy changes. Something would have to spawn a new check_pool thread for the new pool when deploying it, which may add some complication, but is likely possible.

lib/vmpooler/api/v1.rb
```diff
@@ -120,6 +120,74 @@ def atomically_allocate_vms(payload)
...
+      result = { 'ok' => false }
+
+      pool_index = pool_index(pools)
+      pools_updated = 0
+      sync_pool_sizes
+
+      payload.each do |poolname, size|
+        unless pools[pool_index[poolname]]['size'] == size.to_i
+          pools[pool_index[poolname]]['size'] = size.to_i
+          backend.hset('vmpooler__config__poolsize', poolname, size)
```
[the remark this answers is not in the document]
mattkirby: It's because redis always returns a string value for the 'vmpooler__config__poolsize' entry, whether your input was an integer, or not.

docs/API.md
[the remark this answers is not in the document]
ekinanp: Yeah that makes sense, a single endpoint would be nice to have in the future (what's here makes sense and is great too). Do you want me to file a VMPOOLER ticket to consider the single endpoint?

Making the config. file the source of truth definitely makes sense, but from an automation standpoint it would be great if at least a simple iteration of something like adding pools live would go out during the insurgency and only be available on vmpooler-dev (one idea to do this is to introduce a new config. entry 'turn on experimental features: true' that the production instances would not have, but the dev pooler would). What I'm thinking is then the vmpooler-dev. part would consist of (in CI code)

```
(1) Make an API request to add the new pool. You'd know the template and pool size beforehand.
(2) Poll the ready VMs endpoint until size VMs have been moved to Ready. This can fail after 15-20 minutes, since 15-20 minutes is about the time vmpooler stops checking for Ready VMs.
(3) Check-out a ready VM and run some simple smoke tests on it (e.g. ensure you can connect to it as a user, for a start)
(4) Destroy the ready VM via. an API request
(5) Verify that the ready VM has been destroyed
(6) Delete the newly created pool from the config., since it is no longer needed.
(7) Make a PR to puppetlabs-modules adding the new pool to the production instances. Deploying to production can be manual after this part, until we've improved the automation here.
```

I'm thinking the dev pooler wouldn't have any pools, only creating pools when we're about to test an image (and the size for that pool would always be 1).

A single config. endpoint might make this easier (b/c then you can introduce the new pool as part of the config.).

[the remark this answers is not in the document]
mattkirby: It's perhaps a little strange, but couldn't you accomplish this workflow by simply treating a pool title as arbitrary and swapping out its template path? I acknowledge this may seem a little odd, but from vmpooler's perspective I don't think it knows any difference, so in that sense you could have something generic and swap out its template as needed.

Please do create a ticket in the VMPOOLER project for the single endpoint. I'd like to get some input on it and it will help to track the request.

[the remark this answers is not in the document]
ekinanp: Ah yes, we could. Thanks for the suggestion! I think [cricket.pp](https://github.com/puppetlabs/puppetlabs-modules/blob/production/site/profile/manifests/vmpooler/pools/cricket.pp) might need to be updated. I'm thinking of removing all of those pools and just having a single `dev-pool` entry defaulting to a template we know exists w/ size 1. Then we only need to update the template. Would that work/be fine?

Sure, will go ahead and do that.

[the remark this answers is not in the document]
ekinanp: Filed https://tickets.puppetlabs.com/browse/POOLER-110

lib/vmpooler/api/v1.rb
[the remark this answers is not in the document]
ekinanp: Ah that sucks. You could still DRY this up by something like
```
update_pool_config_entry(payload, entry, munge = lambda { |x| x })
```

so that we would have:
```
update_pool_size(payload) = update_pool_config_entry(payload, 'size', lambda { |size| size.to_i })
```
```
update_pool_template(payload) = update_pool_config_entry(payload, 'template')
```

and you can do something similar for the sync. methods.

(The similarity between the update methods is also what lead my Q. about the single `config/pool` endpoint).

lib/vmpooler/api/v1.rb
```diff
@@ -502,6 +572,26 @@ def invalid_templates(payload)
       invalid
     end
 
+    def invalid_template_or_size(payload)
+      invalid = []
+      payload.each do |pool, size|
+        invalid << pool unless pool_exists?(pool)
+        Integer(size) rescue invalid << pool
```
ekinanp: Should this check for `size <= 0`?

docs/API.md
```diff
@@ -540,3 +549,92 @@ $ curl -G -d 'from=2015-03-10' -d 'to=2015-03-11' --url vmpooler.company.com/api
...
+* 200 - OK
+* 404 - No configuration found
+```
+$ curl https://vmpooler.company.com/api/v1/config
+```
+```json
+{
+  "pool_configuration": [
+    {
+      "name": "redhat-7-x86_64",
```
ekinanp: Should this be updated to include the `template_ready` field?

docs/API.md
```diff
@@ -540,3 +549,92 @@ $ curl -G -d 'from=2015-03-10' -d 'to=2015-03-11' --url vmpooler.company.com/api
...
+      "folder": "vmpooler/redhat-7-x86_64",
+      "datastore": "stor1",
+      "size": 1,
+      "datacenter": "dc1",
+      "provider": "vsphere",
+      "capacity": 1,
+      "major": "redhat"
+    }
+  ],
+  "status": {
```
ekinanp: Out of curiosity, why is this under a 'status' key? Unless I'm mistaken, wouldn't that break consistency with the other endpoints?

mattkirby: I was following the convention of the /status endpoint here.

lib/vmpooler/api/v1.rb
mattkirby: Yes, it probably should. I've pushed a change and tests that validate this.

spec/unit/pool_manager_spec.rb
```diff
@@ -1503,6 +1503,393 @@
...
+
+    context 'with a total size greater than the pool size' do
+      let(:ready) { 4 }
+      let(:total) { 4 }
+      it 'should remove excess ready vms' do
+        expect(subject).to receive(:move_vm_queue).exactly(2).times
+
+        subject.remove_excess_vms(config[:pools][0], provider, ready, total)
+      end
+
```
sbeaulie: can you also add my test here to check if it passes?
https://github.com/mattkirby/vmpooler/pull/2/files#diff-04094802b5bab73dcebfdd6e91f2b0a1R1673

lib/vmpooler/pool_manager.rb
```diff
@@ -683,36 +801,53 @@ def _check_pool(pool, provider)
...
+          $redis.del("vmpooler__empty__#{pool['name']}") unless ready.zero?
+        elsif ready.zero?
+          $redis.set("vmpooler__empty__#{pool['name']}", 'true')
+          $logger.log('s', "[!] [#{pool['name']}] is empty")
+        end
+
+        # Check to see if a pool size change has been made via the configuration API
+        # Since check_pool runs in a loop it does not
+        # otherwise identify this change when running
+        update_pool_size(pool)
```
sbeaulie: isn't this called inline in the API method to update the pool size? Why run it in the main loop too?

mattkirby: Because check_pool runs in a loop with the pool configuration passed in it will not otherwise detect this change.

mattkirby:
Agreed on squashing and merging. I can squash by hand first to ensure the commit message is as useful as it can be, unless you think it's all worth keeping in which case I'm all for letting github do it.

mattkirby:
I have squashed this change to a single commit and rebased it on master to include travis changes that allow tests to pass.

Issue #213, opened 2017-03-20
`find_vmdks` in `vsphere_helper` should not use `vmdk_datastore._connection`


glennsarti:
In `find_vmdks` in `vsphere_helper` it uses the call  `vmdk_datastore._connection` to get the underlying connection, however this is already available as `@connection`. It is also using a variable which is not exposed by the vSphere SDK

Instead, it should just use `@connection`
https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/vsphere_helper.rb#L370

PR #260, opened 2018-06-04
Remove propertyCollector from add_disk

mattkirby:
This commit updates add_disk to remove propertyCollector, which was used
to back the find_vmdks method to locate the disk file on datastore and
then use its length to name the new disk. Instead, the number of disks
on the VM is used to ensure a unique disk resource title. Without this
change add_disk can take 10-50x longer due to the propertyCollector
method. Additionally, without this change propertyCollector is used in a
non threadsafe manner, which may cause stability issues for vsphere
provider backends.


mattkirby:
This will resolve #213 by removing find_vmdks.

Issue #207, opened 2017-03-16
Vsphere connections are always insecure (Ignore cert errors)


glennsarti:
Due to the following code the connection to vSphere is always insecure;
https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/vsphere_helper.rb#L29
```
insecure: credentials['insecure'] || true
```
This is because `||` is a binary or operator and anything that is `xxx or true` will evaluate to true.

I imagine the original intent was if `credentials['insecure']` is nil then default to true otherwise use `credentials['insecure']`

mattkirby:
true was hardcoded prior to this, but it looks like this isn't much better.

glennsarti:
insecure is jsut to ignore cert errors.  Not as serious as first thought.
https://github.com/vmware/rbvmomi/blob/master/lib/rbvmomi/vim.rb#L24

PR #261, opened 2018-06-08
Change default vsphere connection behavior

mattkirby:
This commit changes the vsphere connection behavior to set insecure false. Without this change insecure is always set to true when making a connection regardless of the setting provided with the provider configuration.


mattkirby:
This fixes #207.

Issue #262, opened 2018-06-11
create release


logicminds:
Not sure why there has never been an official release.  Be great to start tagging and releasing this software officially instead of always just using master.

mattkirby:
+1

mattkirby:
A release has been tagged and a gem has been published. https://github.com/puppetlabs/vmpooler/pull/288 updates the changelog and version file.

Issue #181, opened 2017-02-08
Feature Request: Add provider support


cdenneen:
Allow vmpooler to not only control vsphere backend but openstack, ec2, etc.
I think the vmpooler model is great, just wish it was available to others not using vsphere.

sschneid:
FWIW, I have this functionality (AWS, GCE) in a [fork](https://github.com/sschneid/vmpooler/tree/ec2_pooling), but it is a bit behind `master` at the moment.  I'm hoping to rebase and clean some things up for a future PR.

glennsarti:
It's in the works...It's just a lot of effort to effectively untangle vSphere.

https://github.com/puppetlabs/vmpooler/pull/179 - Initial Work in progress attempt

https://tickets.puppetlabs.com/browse/POOLER-70
https://tickets.puppetlabs.com/browse/POOLER-73

glennsarti:
@sschneid @cdenneen I've started the process.  Any PRs with POOLER-70 will identify the work that's being done.

glennsarti:
@cdenneen This work has been completed and is now merged.

This ticket can be closed.

cdenneen:
I think the work done has prepped the groundwork but I don't think the
dummy provider would be a useable alternate provider.

glennsarti:
Oh no.  the Dummy Provider is not meant as that.  Do you think some docs on how to add a provider would be sufficient?

cdenneen:
Yes I think many of us might be full cloud shops and don't have vSphere so
providing a way to leverage vmpooler with different backends like AWS, GCE,
Azure, Digital Ocean, openstack, etc will help tremendously. While beaker
supports these hypervisors time it takes to spin these up sometimes so
having a quick pool to pull from and reset when complete I think is great
idea.

mattkirby:
Provider support has been added. You can see the vsphere provider as an example of how this is implemented. The gist is your provider needs to sort out providing the capabilities that pool_manager is trying to leverage and then it should be able to leverage the vmpooler application.

cdenneen:
I think we were waiting for working example of another provider (AWS, GCE, Azure) as the dummy one doesn’t explain how to use this added provider support.

mattkirby:
[...]

PR #263, opened 2018-06-12
Adds a new mechanism to load providers from any gem or file path automatically

logicminds:
This extends #181 by making it easier to load providers.

I am just throwing this up here for initial review.  There are no unit tests at this point but there are also other things that need to be added before this can be merged.  

What this code does is find all gems with providers in 'lib/vmpooler/providers' directories from all gems and loads any file found in that directory. This allows anyone to simply install a new provider gem and start using right away.   I would prefer that we only load the provider that is needed instead of everything.  However, more work will need to be done.   This will make it dead simple to start using external providers since all this is required is too have it installed.  This allows anyone to create a provider and keep separate from main gem.

Currently this loads all providers whenever the vmpooler file is loaded.  I don't think this is optimal and we should probably only load the provider being used or set in the config file.

```
# load all providers including third party plugins
require 'vmpooler/providers'
Vmpooler::Providers.load_all_providers
```

requires #262 (mainly a gemspec file is required)   I had to add a temporary gemspec to make this work.

You can test this code out by doing the following:

```
bundle install
bundle exec pry
1] pry(main)> require 'vmpooler'
Loading file /Users/user1/vmpooler/lib/vmpooler/providers/base.rb
Loading file /Users/user1/vmpooler/lib/vmpooler/providers/dummy.rb
Loading file /Users/user1/vmpooler/lib/vmpooler/providers/vsphere.rb


```

### DIFF
--- a/PROVIDER_API.md
+++ b/PROVIDER_API.md
@@ -1,0 +1,83 @@
+# Provider API
+
+## Create a new provider gem from scratch
+
+### Requirements
+1. the provider code will need to be in lib/vmpooler/providers directory of your gem regardless of your gem name
+2. the main provider code file should be named the same at the name of the provider. ie. (vpshere == lib/vmpooler/providers/vsphere.rb)
+3. The gem must be installed on the same machine as vmpooler
+4. The provider name must be referenced in the vmpooler config file in order for it to be loaded.
+5. Your gem name or repository name should contain vmpooler-<name>-provider so the community can easily search provider plugins
+   for vmpooler.
+### 1. Use bundler to create the provider scaffolding
+
+```
+bundler gem --test=rspec --no-exe --no-ext vmpooler-spoof-provider
+cd vmpooler-providers-spoof/
+mkdir -p ./lib/vmpooler/providers
+cd ./lib/vmpooler/providers
+touch spoof.rb
+
+```
+
+There may be some boilerplate files there were generated, just delete those.
+
+### 2. Create the main provider file
+Ensure the main provider file uses the following code.
+
+
+```ruby
+# lib/vmpooler/providers/spoof.rb
+require 'yaml'
+require 'vmpooler/providers/base'
+
+module Vmpooler
+  class PoolManager
+    class Provider
+      class Spoof < Vmpooler::PoolManager::Provider::Base
+      
+      # at this time it is not documented which methods should be implemented
+      # have a look at the vmpooler/providers/vpshere provider for examples
+
+      end
+
+    end
+  end
+end
+  
+
+```
+
+### 3. Fill out your gemspec
+Ensure you fill out your gemspec file to your specifications.  If you need a dependency please make sure you require them.
+
+`spec.add_dependency "vmware", "~> 1.15"`.
+
+At a minimum you may want to add the vmpooler gem as a dev dependency so you can use it during testing.
+
+`spec.add_dev_dependency "vmpooler", "~> 1.15"`
+
+or in your Gemfile
+
+```ruby
+
+gem 'vmpooler', github: 'puppetlabs/vmpooler'
+```
+
+Also make sure this dependency can be loaded by jruby.  If the dependency cannot be used by jruby don't use it.
+
+### 4. Create some tests
+Your provider code should be tested before releasing.  Copy and refactor some tests from the vmpooler gem under 
+`spec/unit/providers/dummy_spec.rb`
+
+### 5. Publish
+Think your provider gem is good enough for others?  Publish it and tell us on Slack or update this doc with a link to your gem.
+
+
+## Available Third Party Providers
+Be the first to update this list.  Create a provider today!
+
+
+## Example provider
+You can use the following [repo as an example](https://github.com/logicminds/vmpooler-vsphere-provider) of how to setup your provider gem.
+
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -11,7 +11,7 @@ module Vmpooler
   require 'yaml'
   require 'set'
 
-  %w[api graphite logger pool_manager statsd dummy_statsd generic_connection_pool providers].each do |lib|
+  %w[api graphite logger pool_manager statsd dummy_statsd generic_connection_pool].each do |lib|
     require "vmpooler/#{lib}"
   end
 
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -1,3 +1,5 @@
+require 'vmpooler/providers'
+
 module Vmpooler
   class PoolManager
     CHECK_LOOP_DELAY_MIN_DEFAULT = 5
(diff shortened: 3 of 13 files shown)


logicminds:
@glennsarti

glennsarti:
@logicminds Heya Corey... You just want me to review?

logicminds:
Thoughts on using this approach?  I use this type of design for retrospec, puppet-debugger and a few other places.  

Questions:
1. Why is a provider also titled pool_manager?
2. Does lib/pool_manager have anything to do with lib/providers?
3. Should we load the providers from the pool_manager.rb?
4. Can the config have more than one provider at a time?  (Is this a future feature?)
5. Where and should we load/use the providers?  Seems like we could be lazier when loading things.

lib/vmpooler/pool_manager.rb
```diff
@@ -26,6 +28,9 @@ def initialize(config, logger, redis, metrics)
       @reconfigure_pool = {}
 
       @vm_mutex = {}
+
+      # load specified providers from config file
+      load_used_providers
```
logicminds: All the providers referenced in the config are now loaded here.  Previously these were loaded during the loading of the vmpooler file.

logicminds:
@glennsarti This is ready to use now unless anyone has some concerns.

logicminds:
I updated the dockerfile so I could find this failure.  I never found the failure but looks like my change fixed the tests anyway.

mattkirby:
The Dockerfile is changing significantly in #281 . I think the first failure may have been erroneous. Can you remove the Dockerfile changes since they appear otherwise unrelated?

logicminds:
@mattkirby 

```
 however if they are to be loaded as a separate gem then I think there should be no requirement of them being in the providers directory since when the gem is published there will be no need to run vmpooler from source any longer
```
In case it was not clear.  The providers themselves don't need to be in this repository.  But any provider plugin will need to have the same lib/vmpooler/providers directory layout in order for this mechanism to discover it.  Otherwise, there is no good way to determine if a plugin is for vmpooler. File based detection is the most efficient way AFAIK. Additionally, if the plugin is supposed to have the namespace of `Vmpooler::Providers` then it is good ruby practice to put the file in a similar named directory `lib/vmpooler/providers/`.  See #284 for why I don't like the current layout though.

I removed my dockerfile mods.

mattkirby:
@logicminds thanks for the clarification, I think I understand now.

logicminds:
I'll create a example provider repo just in case and attach to my PROVIDER doc.

logicminds:
Here is a example gem provider.  https://github.com/logicminds/vmpooler-vsphere-provider

You could also use this as a foundation for the vsphere provider and them remove that provider from this repo.

lib/vmpooler.rb
```diff
@@ -13,7 +14,9 @@ module Vmpooler
   require 'yaml'
   require 'set'
 
-  %w[api graphite logger pool_manager statsd dummy_statsd generic_connection_pool providers].each do |lib|
+
```
mattkirby: Nitpick: extra lines added here.

logicminds:
Need to drop support for ruby 2.2 in travis as it is failing my tests.  Gemspec says `ruby >= 2.3`.

@mattkirby do you want me to remove 2.2 from travis file?

mattkirby:
I put up https://github.com/puppetlabs/vmpooler/pull/286 to remove that. I forgot to get that added to the branch that just landed, sorry.

logicminds:
I'll rebase once 285, 286 and 287 are merged.

PROVIDER_API.md
```diff
@@ -0,0 +1,83 @@
+# Provider API
+
+## Create a new provider gem from scratch
+
+### Requirements
+1. the provider code will need to be in lib/vmpoolers/providers directory of your gem regardless of your gem name
+2. the main provider code file should be named the same at the name of the provider. ie. (vpshere == lib/vmpoolers/providers/vsphere.rb)
+3. The gem must be installed on the same machine as vmpooler
+4. The provider name must be referenced in the vmpooler config file in order for it to be loaded.
+5. Your gem name or repository name should contain vmppoler-<name>-provider so the community can easily search provider plugins
```
glennsarti: nit: Typo `vmppoler1`

spec/unit/pool_manager_spec.rb
```diff
@@ -46,6 +46,30 @@
...
+:providers:
+  :mock:
+:pools:
+  - name: '#{pool}'
+    size: 1
+    provider: 'spoof'
+    EOT
+    )
+    }
+    it do
```
glennsarti: nit: Would prefer a descriptive name for test

mattkirby: +1 to descriptive name for test, otherwise looks ready to roll.

lib/vmpooler/providers.rb
```diff
@@ -1,7 +1,116 @@
...
+    # @return [Array] returns a list of vmpooler providers gem plugin specs
+    def vmpooler_provider_gem_list
+      gemspecs.find_all { |spec| File.directory?(File.join(spec.full_gem_path, provider_path)) } + included_lib_dirs
+    end
+
+    # Internal: Find any gems containing vmpooler provider plugins and load the main file in them.
+    #
+    # @return [Array[String]] - a array of provider files
+    # @param name [String] - the name of the provider to load
+    def load_from_gems(name = nil)
```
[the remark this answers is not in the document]
logicminds: This already exists.  But the code is in the pool_manager.rb file instead because I need to separate the concerns.  The code below is what currently resides in this MR for pool_manager.

```
# load only providers used in config file
    def load_used_providers
      Vmpooler::Providers.load_by_name(used_providers)
    end

    # @return [Array] - a list of used providers from the config file, defaults to the default providers
    # ie. ["vsphere", "dummy"]
    def used_providers
      pools = config[:pools] || []
      @used_providers ||= (pools.map { |pool| pool[:provider] || pool['provider'] }.compact + default_providers ).uniq
    end

    # @param [Array] - returns a list of providers that should always be loaded
    # note: vsphere is the default if user does not specify although this should not be
    # if vsphere is to no longer be loaded by default please remove
    def default_providers
      @default_providers ||= %w( vsphere dummy )
    end
```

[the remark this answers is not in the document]
mattkirby: Thanks @logicminds . I'm good with merging this, just have one last nitpick on the test name that is untitled and otherwise this looks ready to roll.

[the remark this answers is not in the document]
logicminds: Pretty sure I addressed that already.

```
it '#correct class' do
    expect(providers).to be_a Vmpooler::Providers
end
```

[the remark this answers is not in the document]
mattkirby: In the providers spec I see your example, but the one I was referring to is the one with an in-line comment here https://github.com/puppetlabs/vmpooler/pull/263/files#diff-04094802b5bab73dcebfdd6e91f2b0a1R62

PR #265, opened 2018-06-13
(POOLER-112) Ensure a VM is only destroyed once

mattkirby:
This commit implements a vm_mutex hash to allow synchronizing VM operations that should only happen once across threads. Without this change pool_manager will try to evaluate or destroy a VM multiple times, which results in an error being thrown by one of the destroy attempts as only one can succeed and a duplication of resources unnecessarily when there are no errors.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -24,6 +24,8 @@ module Vmpooler
 
       # Pool mutex
       @reconfigure_pool = {}
+
+      @vm_mutex = {}
     end
 
     def config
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -555,6 +577,10 @@ def check_pool(pool,
       end
     end
 
+    def vm_mutex(vmname)
+      @vm_mutex[vmname] || @vm_mutex[vmname] = Mutex.new
```
sbeaulie: will this grow forever or are the old mutex objects garbage collected at one point?

mattkirby: From the research I've done about this ruby will clean up these objects as a part of the garbage collection process so long as they are no longer being used. Given the lifecycle of a VM I expect that these objects should not be around longer than the lifecycle of a given VM, which will allow ruby to clean them up.

sbeaulie: I talking specifically about the hash `vm_mutex`. Can we have a look throught the lifecycle of a few VMs and keep an eye on that hash to see if it only grows?

mattkirby: Yes. That's where we will build up several hundred mutex objects. I did perform a test of creating a few hundred VMs, watching memory usage increased, destroy the VMs, and then watch memory usage return to the same value as startup time (over time, not right away, quickly reduced to 80%, after a few hours was back to startup time value). I think this is an indication that the object dereferencing and garbage collection are working as expected, but I'm happy to consider how we can test this further.

mattkirby:
I had to add one more small change here. Migrations will wait for the lock to free. Without this addition the running VM check would always cause the mutex to be locked and cause migrations to fail.

PR #266, opened 2018-06-21
(POOLER-31) Expire redis vm key when clone fails

mattkirby:
This commit updates pool_manager to expire a redis VM key when a clone fails. Without this change VMs that fail to clone have their metadata left forever.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -245,7 +245,9 @@ module Vmpooler
 
         $metrics.timing("clone.#{pool_name}", finish)
       rescue => _err
-        $redis.srem('vmpooler__pending__' + pool_name, new_vmname)
+        $redis.srem("vmpooler__pending__#{pool_name}", new_vmname)
+        expiration_ttl = $config[:redis]['data_ttl'].to_i * 60 * 60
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -245,7 +245,9 @@ def _clone_vm(pool, provider)
 
         $metrics.timing("clone.#{pool_name}", finish)
       rescue => _err
-        $redis.srem('vmpooler__pending__' + pool_name, new_vmname)
+        $redis.srem("vmpooler__pending__#{pool_name}", new_vmname)
+        expiration_ttl = $config[:redis]['data_ttl'].to_i * 60 * 60
```
Dakta: It looks like redis is taking [seconds](https://redis.io/commands/expire) for this value, so `data_ttl` is in units of hours? Can this be documented somewhere?

mattkirby: It looks like there is a mention of this in the example configuration file https://github.com/puppetlabs/vmpooler/blob/master/vmpooler.yaml.example#L180 .

Dakta: 👍

PR #267, opened 2018-06-26
(POOLER-113) Add support for multiple LDAP search bases

mattkirby:
This commit updates vmpooler to support setting an array of search bases
in addition to a single base provided as a string. Without this change
it is not possible to specify multiple search bases to use with the LDAP
authentication provider. Additionally, test coverage is added to
the authentication helper method.

### DIFF
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -54,35 +54,63 @@ module Vmpooler
         return false
       end
 
+      def authenticate_ldap(port, host, user_object, base, username_str, password_str)
+        require 'rubygems'
+        require 'net/ldap'
+
+        ldap = Net::LDAP.new(
+          :host => host,
+          :port => port,
+          :encryption => {
+            :method => :start_tls,
+            :tls_options => { :ssl_version => 'TLSv1' }
+          },
+          :base => base,
+          :auth => {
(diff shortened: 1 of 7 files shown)


sbeaulie:
The new LDAP tests did not pass in travis

mattkirby:
Interesting. Those tests work if I test the file they're in alone. However, I get the same failures when running the full suite. I'll investigate and update this.

mattkirby:
It looks like tests were failing because `authenticate` was redefined in 'token_spec.rb'. I've updated this to remove the duplicate definition of `authenticate` and updated 'token_spec.rb' to use the dummy auth provider.

lib/vmpooler/api/helpers.rb
```diff
@@ -54,35 +54,63 @@ def authorized?
         return false
       end
 
+      def authenticate_ldap(port, host, user_object, base, username_str, password_str)
+        require 'rubygems'
+        require 'net/ldap'
```
mchllweeks: @mattkirby Just curious why you are putting the require statement inside the authenticate_ldap method instead of at the top of the file.  Is there a reason to limit the scope?

mattkirby: Only because it's what was already being done. It probably should be at the top of the file.

Issue #204, opened 2017-03-14
`find_folder` in vsphere_helper.rb has subtle errors


glennsarti:
In https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/vsphere_helper.rb#L174-L178 `find_folder` is only capable of folders in the vmFolder hierarchy as per https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.Datacenter.html .  This method should be renamed to `find_vm_folder` or something similar to more properly describe what it does.

Also, the search method does not handle VMs or VM templates with the same name as folders (which is perfectly valid).  The name check should also determine what type of object it s before it decides to use it.

PR #268, opened 2018-06-28
Reduce object lookups for finding folders

mattkirby:
These commits reduce the number of objects looked up when finding VM folders. Additionally, object lookup methods that scan for a folder with a matching name are removed in favor of searchIndex with FindByInventoryPath. Without this change the time it takes to find VM folders is much longer and more resource intensive.

### DIFF
--- a/lib/vmpooler/providers/vsphere.rb
+++ b/lib/vmpooler/providers/vsphere.rb
@@ -46,8 +46,7 @@ module Vmpooler
           vms = []
           @connection_pool.with_metrics do |pool_object|
             connection = ensured_vsphere_connection(pool_object)
-            foldername = pool_config(pool_name)['folder']
-            folder_object = find_folder(pool_name, connection)
+            folder_object = find_vm_folder(pool_name, connection)
 
             return vms if folder_object.nil?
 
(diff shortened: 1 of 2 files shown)


mattkirby:
This will resolve https://github.com/puppetlabs/vmpooler/issues/204 .

lib/vmpooler/providers/vsphere.rb
```diff
@@ -365,15 +353,22 @@ def get_target_datacenter_from_config(pool_name)
           nil
         end
 
-        def generate_vm_hash(vm_object, template_name, pool_name)
-          hash = { 'name' => nil, 'hostname' => nil, 'template' => nil, 'poolname' => nil, 'boottime' => nil, 'powerstate' => nil }
+        def generate_vm_hash(vm_object, pool_name)
```
Dakta: This seems like a good opportunity to add documentation to this method.

mattkirby: I added some brief documentation to this method. Let me know if you think it doesn't cover what you were hoping to see, it's pretty brief.

Dakta: 👍

PR #271, opened 2018-07-02
Ensure template deltas are created once

mattkirby:
This commit updates how template delta disk creation is evaluated. Without this change template deltas are created for every template on each applicatoin startup. This change updates this behavior to instead run template delta disk creation only once per template configured for a pool. Without this change it is possible to get a template to a state where the XML depth is too great to be read with default settings and the template requires a new clone to resolve.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -628,6 +628,11 @@ module Vmpooler
           prepare_template(pool, provider)
           prepared_template = $redis.hget('vmpooler__template__prepared', pool['name'])
         end
+      elsif not prepared_template == pool['template']
+        mutex.synchronize do
+          prepare_template(pool, provider)
+          prepared_template = $redis.hget('vmpooler__template__prepared', pool['name'])
+        end
       end
       return if configured_template.nil?
       return if configured_template == prepared_template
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -628,6 +628,11 @@ def evaluate_template(pool, provider)
           prepare_template(pool, provider)
           prepared_template = $redis.hget('vmpooler__template__prepared', pool['name'])
         end
+      elsif not prepared_template == pool['template']
```
kevpl: @mattkirby in general, I would advocate for not using the "word" form of boolean operators, because they make statements harder to reason about due to [precendence order](https://stackoverflow.com/a/27429722).

This seems like a simple enough case that I wouldn't worry about it in particular, but when we get to more complex statements, it could get ugly. I would suggest we implement a standard for ourselves to unify around this. Searching for other internal ruby standards, I'm seeing a few projects of ours ([higgs](https://github.com/puppetlabs/higgs/blob/2018.2.0-rc0/README.md#code-style), [courseware](https://github.com/puppetlabs/courseware/blob/master/_archive/extending/Introduction/style_guide_tools.md#style-tools)) pointing to [github's ruby styleguide](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md), which has this opinion about the other operators of this kind, which I think would apply to `not` as well:

> The and and or keywords are banned. It's just not worth it. Always use && and || instead.

my favorite example I've seen so far against this practice would be [this one](https://github.com/rubocop-hq/ruby-style-guide/issues/329#issuecomment-62769404).

mattkirby: I wasn't sure when using this operator initially whether it was a good idea so it's good to see some ways in which it's not a good idea. I've updated usage of this to instead be `elsif !=`. Let me know if you think that should be changed further.

PR #272, opened 2018-07-03
(POOLER-40) Do not return folders with get_pool_vms

mattkirby:
This commit updates get_pool_vms to only return if the object is a VirtualMachine. Without this change a folder wtihin a VM target folder can be discovered and destroyed as though it were a VM.

### DIFF
--- a/spec/unit/providers/vsphere_spec.rb
+++ b/spec/unit/providers/vsphere_spec.rb
@@ -159,6 +159,30 @@ EOT
         expect(result).to eq(expected_vm_list)
       end
     end
+
+    context 'given a pool folder with a folder and vms' do
+      let(:expected_vm_list) {[
+        { 'name' => 'vm1'},
+        { 'name' => 'vm2'},
+        { 'name' => 'vm3'}
+      ]}
+      let(:folder_object2) { mock_RbVmomi_VIM_Folder({ :name => 'pool2'}) }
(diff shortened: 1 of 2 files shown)


spec/unit/providers/vsphere_spec.rb
```diff
@@ -159,6 +159,30 @@
...
+          mock_vm = mock_RbVmomi_VIM_VirtualMachine({ :name => vm_hash['name'] })
+          # Add the mocked VM to the folder
+          folder_object.childEntity << mock_vm
+        end
+        folder_object.childEntity << folder_object2
+        expect(subject).to receive(:find_folder).with(pool_config['folder'],connection,datacenter_name).and_return(folder_object)
+      end
+
+      it 'should return the vms without the folder' do
+        result = subject.vms_in_pool(poolname)
```
sbeaulie: that looks good, do you want it to also return VMs inside the folder 'pool2' or is it supposed to skip these?

mattkirby: No, it's supposed to skip those and just evaluate the objects that are VMs in the base folder object. Previously it would treat the folder as though it was a VM and try to discover and destroy it.

sbeaulie: Yes I've seen that before. I wonder why we would have subfolders though

PR #274, opened 2018-07-06
(POOLER-66) Purge vms and folders no longer configured

mattkirby:
This commit enables optional purging for vms and folders when they are
not configured in the vmpooler provided configuration. Base folders are
determined from folders specified in the pool configuration. Then,
anything not configured in that folder for that provider and is not a
whitelisted folder title will be destroyed. Without this change vmpooler
will leave unconfigured vms and folders behind and any vms will be left
running forever without manual intervention. Additionally, any
associated redis data will never be expired.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -287,6 +287,54 @@ module Vmpooler
       end
     end
 
+    def purge_unused_vms_and_folders
+      global_purge = $config[:config]['purge_unconfigured_folders']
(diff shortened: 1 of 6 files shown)


mattkirby:
I added one more change that I found was necessary, referencing the provider_config object instead of trying to get the provider name again and finding its config via the global config object. I had not previously realized I could access the provider configuration this way.

sbeaulie:
Can you add the new config options to the documentation?

mattkirby:
Yes, good point.

PR #276, opened 2018-07-06
(POOLER-81) Add time_remaining information

smcelmurry:
Before, the only time calculation displayed for a given VM was the
lifetime parameter. Added the time_remaining parameter which will
display time until the VM is destroyed in hours, minutes, seconds.

Additionally, updated the running parameter to display in a similar
fashion as time_remaining.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -644,7 +644,10 @@ module Vmpooler
           result[params[:hostname]]['running'] = ((Time.parse(rdata['destroy']) - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
           result[params[:hostname]]['state'] = 'destroyed'
         elsif rdata['checkout']
-          result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          running = Time.now - Time.parse(rdata['checkout'])
+          result[params[:hostname]]['running'] = format("%02dh %02dm %02ds", running / (60 * 60), (running / 60) % 60, running % 60)
+          remaining = Time.parse(rdata['checkout']) + rdata['lifetime'].to_i*60*60 - Time.now
+          result[params[:hostname]]['time_remaining'] = format("%02dh %02dm %02ds", remaining / (60 * 60), (remaining / 60) % 60, remaining % 60)
           result[params[:hostname]]['state'] = 'running'
         elsif rdata['check']
           result[params[:hostname]]['state'] = 'ready'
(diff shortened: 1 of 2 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -644,7 +644,10 @@ def invalid_template_or_path(payload)
           result[params[:hostname]]['running'] = ((Time.parse(rdata['destroy']) - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
           result[params[:hostname]]['state'] = 'destroyed'
         elsif rdata['checkout']
-          result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          running = Time.now - Time.parse(rdata['checkout'])
+          result[params[:hostname]]['running'] = format("%02dh %02dm %02ds", running / (60 * 60), (running / 60) % 60, running % 60)
```
sbeaulie: modulos! modulo

sbeaulie: My only question would be if something is using this existing API, like kermi or bitbar/vmfloaty?

smcelmurry: I tested this with vmfloaty and it just returns the json:
```
➜  vmpooler git:(pooler-81) # floaty get debian-7-i386
{"debian-7-i386":"poolvm-me51df2cadymlq0.company.com"}
➜  vmpooler git:(pooler-81) # floaty query poolvm-me51df2cadymlq0
     {
      "ok"=>true,
      "poolvm-me51df2cadymlq0"=>
       {"template"=>"debian-7-i386",
        "lifetime"=>24,
        "running"=>"00h 00m 08s",
        "time_remaining"=>"23h 59m 51s",
        "state"=>"running",
        "ip"=>"",
        "domain"=>"company.com"}
     }
➜  vmpooler git:(pooler-81) #
```
Not sure about bitbar or kerminator. Is there a plugin for kermi? I wasn't aware of that

mattkirby: It looks like bitbar expects this to be a float64 value. So, we may consider pushing an update there before rolling out this change. Either way I think it's reasonable to merge and we can discuss whether to update bitbar before pushing this change out to instances.

johnmccabe: I'd personally prefer if the running/time_remaining returned from the API didn't come formatted, would be preferable to leave this up to the client.

glennsarti: As this is a public API, this would be considered a breaking change (going from a Float to Formatted String).  I know my PowerShell Pooler code will blow up because of this.

+1 to @johnmccabe .  If the data is coming back formatted, the API should expose the raw data somehow and/or the explicitly say this is the "friendly" text version e.g.

```
     {
      "ok"=>true,
      "poolvm-me51df2cadymlq0"=>
       {"template"=>"debian-7-i386",
        "lifetime"=>24,
        "running"=>"0.01",
        "start_time" => "2018-01-01 01:02:03Z"
        "time_remaining"=>"23.99",
        "end_time" => "2018-01-02 01:02:03Z"
        "state"=>"running",
        "ip"=>"",
        "domain"=>"company.com"}
     }
```

PR #277, opened 2018-07-06
(POOLER-34) Ship clone request to ready time to metrics

smcelmurry:
Before, we were already capturing this metric but we failed to ship
it anywhere. This ships the appropriate metric as `time_to_ready_state`
Dashboards can be created easily in grafana.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -102,7 +102,7 @@ module Vmpooler
 
         # last boot time is displayed in API, and used by alarming script
         $redis.hset('vmpooler__lastboot', pool, Time.now)
-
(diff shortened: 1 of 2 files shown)


smcelmurry:
@mattkirby It didn't look like any spec tests needed to change for this since we are only testing if metrics are generally getting sent or not.

sbeaulie:
Great! I know John O'Connor was looking to do some analysis of ready times for improvements to the windows images (like running sysprep part of the build)

mattkirby:
Looks reasonable to me. We could possibly add some test coverage for this addition by ensuring that metrics receives the expected timing when testing the method here https://github.com/puppetlabs/vmpooler/blob/27c9ee75912a96377daaa48ecf4e0821d3d0538a/spec/unit/pool_manager_spec.rb#L182 .

PR #280, opened 2018-07-10
(POOLER-81) Add time remaining information

smcelmurry:
Before, the only time calculation displayed for a given VM was the
lifetime parameter. Added the remaining parameter which will
display time until the VM is destroyed as a float.

Additionally, start_time and end_time were added to api to return
as RFC3339 based times (e.g. 2018-07-11T09:54:51-07:00).

/cc @glennsarti @johnmccabe @mattkirby

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -646,8 +646,8 @@ module Vmpooler
         elsif rdata['checkout']
           result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
           result[params[:hostname]]['remaining'] = ((Time.parse(rdata['checkout']) + rdata['lifetime'].to_i*60*60 - Time.now) / 60 / 60).round(2)
-          result[params[:hostname]]['start_time'] = Time.parse(rdata['checkout'])
-          result[params[:hostname]]['end_time'] = Time.parse(rdata['checkout']) + rdata['lifetime'].to_i*60*60
+          result[params[:hostname]]['start_time'] = Time.parse(rdata['checkout']).to_datetime.rfc3339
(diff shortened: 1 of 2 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -645,6 +645,9 @@ def invalid_template_or_path(payload)
           result[params[:hostname]]['state'] = 'destroyed'
         elsif rdata['checkout']
           result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          result[params[:hostname]]['remaining'] = ((Time.parse(rdata['checkout']) + rdata['lifetime'].to_i*60*60 - Time.now) / 60 / 60).round(2).abs
```
smcelmurry: `abs` is used here to avoid a flaky test in `vm_spec.rb`. `remaining` would sometimes return as a negative.

lib/vmpooler/api/v1.rb
```diff
@@ -645,6 +645,9 @@ def invalid_template_or_path(payload)
           result[params[:hostname]]['state'] = 'destroyed'
         elsif rdata['checkout']
           result[params[:hostname]]['running'] = ((Time.now - Time.parse(rdata['checkout'])) / 60 / 60).round(2)
+          result[params[:hostname]]['remaining'] = ((Time.parse(rdata['checkout']) + rdata['lifetime'].to_i*60*60 - Time.now) / 60 / 60).round(2)
+          result[params[:hostname]]['start_time'] = Time.parse(rdata['checkout'])
```
glennsarti: Personally, I'd like to see an explicit format chosen here, instead of at the whim of Ruby. i.e. is is UTC or does it include the originating timezone.

johnmccabe: I'd love to see RFC3339 here

glennsarti: 👍  to RFC3339

smcelmurry: Updated!

Issue #209, opened 2017-03-16
`find_pool` in vsphere_helper tends to throw instead of returning nil for missing pools


glennsarti:
In `find_pool` in `lib/vmpooler/vsphere_helper.rb` if the poolname is not found the base object turns into `nil` (Assuming it's on the same hierarchy).  However due to https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/vsphere_helper.rb#L292

``` ruby
 base = base.resourcePool unless base.is_a?(RbVmomi::VIM::ResourcePool) && base.respond_to?(:resourcePool)
```
If `base == nil` then the unless is not trigged and and an error is thrown that `resourcePool` is a missing method on `nil`

mattkirby:
Find_pool is no longer used. I will remove it.

Issue #210, opened 2017-03-17
`find_pool` in vsphere_helper.rb has subtle errors


glennsarti:
The logic used to traverse the tree uses the object names for traversal, however if two objects have same name e.g. a Folder and a Resource Pool, and are in a particular order find_pool will throw errors.

This is similar to issue https://github.com/puppetlabs/vmpooler/issues/204 .

Perhaps the find_pool code should be using a PropertyCollector to get vCenter to do an inventory search instead.

e.g. http://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.wssdk.pg.doc%2FPG_PropertyCollector.7.1.html

mattkirby:
Find_pool is no longer used. I will submit a PR to remove this.

PR #283, opened 2018-07-13
(MAINT) Remove find_pool and update pending tests

mattkirby:
This commit removes find_pool, which is no longer used. Additionally, tests which are listed as pending and have been resolved are removed from pending tests. Lastly, a folder check that doesn't make sense given the change in method for finding objects is removed.


mattkirby:
This will resolved #209 and #210

mattkirby:
This change reduces pending checks to three!

PR #295, opened 2018-07-31
Bring the pitchfork

kevpl:
This PR is a result of the mobbing process that QE has done together over the course of the last month.

Note that we don't want to merge this PR as-is, the git history is crazy, but we at least wanted to have the original git history available for review for those (@mattkirby) who didn't get to see what happened on a particular day.

One issue that we'd like your input on specifically @mattkirby is that there is a test that involves a query on a redis key with the value '*__updating__*'. It looks like any use of this has been removed & replaced with the mutex locks, is that correct?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+# For local development run with a dummy provider
+version: '3'
+services:
+  aio-local:
+    build:
+      context: .
+      dockerfile: Dockerfile-aio
+    volumes:
+      - ".:/var/lib/vmpooler"
+    ports:
+      - "80:4567"
+    environment:
+      - VMPOOLER_DEBUG=true # for use of dummy auth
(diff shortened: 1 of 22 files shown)


mattkirby:
Yes, the 'updating' redis key usage has been removed in favor of using mutex objects. This is some cleanup of test coverage that I had implemented for a prior implementation and failed to rip out.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+# For local development run with a dummy provider
+version: '3'
+services:
+  aio-local:
+    build:
+      context: .
+      dockerfile: Dockerfile-aio
+    volumes:
+      - ".:/var/lib/vmpooler"
```
mattkirby: We no longer need to map in this directory. Instead we should consider mapping in a configuration file to the default location, or specify one and map a file to that location, similarly to how we specify the configuration to marathon.

PR #298, opened 2018-09-12
(POOLER-129) Allow setting weights for backends

mattkirby:
This commit updates get_vm in the vmpooler API to allow for setting weights for backends. Additionally, when an alias for a pool exists, and the backend configured is not weighted, then the selection of the pool based on alias will be randomly sampled. Without this change any pool with the title of the alias is exhausted before an alternate pool with the configured alias is used, which results in an uneven distribution of VMs. When all backends involved are configured with weighted values the VM selection will be based on probability using those weights.

A bug is fixed when setting the default ttl for check_ready_vm.

Pickup is added to handle weighted VM selection.

A dockerfile is added that allows for building and installing vmpooler
from the current HEAD in docker to make for easy testing.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -36,16 +36,46 @@ module Vmpooler
       validate_token(backend)
     end
 
-    def fetch_single_vm(template)
-      vm = backend.spop('vmpooler__ready__' + template)
-      return [vm, template] if vm
+    def pool_backend(pool)
+      pool_index = pool_index(pools)
+      pool_config = pools[pool_index[pool]]
+      backend = pool_config['clone_target'] || config['clone_target'] || 'default'
+      return backend
+    end
 
+    def fetch_single_vm(template)
+      template_backends = [template]
       aliases = Vmpooler::API.settings.config[:alias]
-      if aliases && aliased_template = aliases[template]
-        vm = backend.spop('vmpooler__ready__' + aliased_template)
-        return [vm, aliased_template] if vm
+      if aliases
+        template_backends << aliases[template] if aliases[template]
+
+        pool_index = pool_index(pools)
+        weighted_pools = {}
+        template_backends.each do |t|
+          next unless pool_index.key? t
+          index = pool_index[t]
+          clone_target = pools[index]['clone_target'] || config['clone_target']
+          next unless config.key?('backend_weight')
+          weight = config['backend_weight'][clone_target]
+          if weight
+            weighted_pools[t] = weight
+          end
+        end
+
+        if weighted_pools.count == template_backends.count
+          pickup = Pickup.new(weighted_pools)
+          selection = pickup.pick
+          template_backends.delete(selection)
+          template_backends.unshift(selection)
+        else
+          template_backends = template_backends.sample(template_backends.count)
+        end
       end
 
+      template_backends.each do |t|
+        vm = backend.spop('vmpooler__ready__' + t)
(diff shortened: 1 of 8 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -36,16 +36,46 @@ def need_token!
...
+      if aliases
+        template_backends << aliases[template] if aliases[template]
+
+        pool_index = pool_index(pools)
+        weighted_pools = {}
+        template_backends.each do |t|
+          next unless pool_index.key? t
+          index = pool_index[t]
+          clone_target = pools[index]['clone_target'] || config['clone_target']
+          next unless config.key?('backend_weight')
```
sbeaulie: We should add this `backend_weight` to the docs.

lib/vmpooler/providers/vsphere.rb
```diff
@@ -428,7 +428,7 @@ def destroy_vm(pool_name, vm_name)
 
         def vm_ready?(_pool_name, vm_name)
           begin
-            open_socket(vm_name)
+            open_socket(vm_name, global_config[:config]['domain'])
```
sbeaulie: this looks like an unrelated change?

lib/vmpooler/api/v1.rb
```diff
@@ -36,16 +36,46 @@ def need_token!
       validate_token(backend)
     end
 
-    def fetch_single_vm(template)
-      vm = backend.spop('vmpooler__ready__' + template)
-      return [vm, template] if vm
+    def pool_backend(pool)
```
sbeaulie: a similar method already exists in the helpers and used 4-5 times in the code

https://github.com/puppetlabs/vmpooler/blob/2daa5244b8f71d3ceecf6695c235e3f2b0c30dbe/lib/vmpooler/providers/vsphere.rb#L451-L459

lib/vmpooler/providers/vsphere.rb
mattkirby: Yes. I previously implemented the ability to specify a VM domain to use when opening a socket connection. Without this change that domain setting is never used. When interacting with a local instance of vmpooler in docker I found that I could not get VMs to reach the ready state without this addition. I could make this a separate commit if you think that makes more sense.

lib/vmpooler/api/v1.rb
```diff
@@ -36,16 +36,46 @@ def need_token!
...
+          weight = config['backend_weight'][clone_target]
+          if weight
+            weighted_pools[t] = weight
+          end
+        end
+
+        if weighted_pools.count == template_backends.count
+          pickup = Pickup.new(weighted_pools)
+          selection = pickup.pick
+          template_backends.delete(selection)
```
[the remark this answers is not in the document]
sbeaulie: I think I'm following the logic now. Why not capture the selection here and in the `else` below. Then instead of doing `template_backends.each` on line 75 again, just `backend.spop()` the selection?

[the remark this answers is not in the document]
mattkirby: My goal is to order the array so the first VM we try to retrieve is based on the selection from pickup. There is an each statement because we need to try this for the next one in the list in case there are no ready VMs for the first selection.

mattkirby:
From what I understand with reading the pickup docs you are setting a probability with that value that is relative to the other values, but has no requirements for equalling a total value, or anything like that. Do you think the example would be more clear if single digits, or values that do not total 100 were used?

PR #312, opened 2019-01-18
Stop testing ruby 2.3.x

mattkirby:
This commit updates travis configuration to stop testing 2.3.x and jruby 9.1.x. Without this change tests fail due to library incompatibilities with ruby 2.3.x.


mattkirby:
This will get tests passing on master again.

PR #314, opened 2019-01-30
(POOLER-138) Support multiple pools per alias

mattkirby:
This change updates handling of pool aliases to allow for more than a
single pool to be configured as an alias pool. Without this change if
multiple pools are configured as an alias the last one to configure it
is set as the alias for that pool.

Additionally, redis testing requirements are removed in favor of
mock_redis. Without this change a redis server is required to run
vmpooler tests.

### DIFF
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -108,10 +108,11 @@ module Vmpooler
       parsed_config[:pool_names] << pool['name']
       if pool['alias']
         if pool['alias'].is_a?(Array)
-          pool['alias'].each do |a|
+          pool['alias'].each do |pool_alias|
             parsed_config[:alias] ||= {}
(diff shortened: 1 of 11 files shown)


lib/vmpooler.rb
```diff
@@ -110,7 +110,8 @@ def self.config(filepath = 'vmpooler.yaml')
         if pool['alias'].is_a?(Array)
           pool['alias'].each do |a|
```
mchllweeks: This is a nitpick, but could you change the variable 'a' to a more descriptive name?

mattkirby: Sure thing. I've updated the variable name.

mattkirby:
Thanks. I've changed the variable name to something more descriptive.

PR #316, opened 2019-02-04
Fix Dockerfile link in readme and add note about http requests for dev

briancain:
This pull request fixes up a dead link in the README to the Dockerfile, and also leaves a note in the dummy example config about making HTTP requests locally.


mattkirby:
@briancain this looks reasonable. Can you please change the commit title to something more descriptive?

briancain:
@mattkirby sounds good. updated the pull request title to match the two commits. Thanks!

PR #319, opened 2019-03-08
Improve cluster lookup

trlinkin:
Without this commit, cluster lookup was done by traversing the first
level below a given datacenter and comparing the names of the objects
present there. Though uncommon, entities at this level may be folders
that contain the sought after cluster. To make matters worse, a non
cluster entity may have the exact same name as the cluster being looked
up. Matching is performed by comparing only entity name and thus a
folder entity may be mistakenly returned if it matches the name provided
for lookup. Traversal into the folder entity is impossible. There is no
error checking or helpful output to determine the issue.

This commit changes the method by which cluster lookup is performed. It
assumes that, as before, the cluster name will be provided as a string
and it will be looked up by that string. However the string can be a
relative entity path as well used to traverse a folder entity. The new
lookup uses the #traverse() method from the datacenter entity to find
the cluster entity. There is also some simple return type checking to
tell if what we found in the traversal is a
`RbVmomi::VIM::ClusterComputeResource` object. If not, a basic
indication of the problem is raised to the user.

### DIFF
--- a/lib/vmpooler/providers/vsphere.rb
+++ b/lib/vmpooler/providers/vsphere.rb
@@ -839,7 +839,9 @@ module Vmpooler
         def find_cluster(cluster, connection, datacentername)
           datacenter = connection.serviceInstance.find_datacenter(datacentername)
           raise("Datacenter #{datacentername} does not exist") if datacenter.nil?
-          datacenter.hostFolder.children.find { |cluster_object| cluster_object.name == cluster }
+          found_cluster = datacenter.hostFolder.traverse(cluster)
+          raise("Cluster #{cluster} could not be found or does not exist") if not found_cluster.is_a? RbVmomi::VIM::ClusterComputeResource
+          found_cluster
         end
 
         def get_cluster_host_utilization(cluster, model = nil)


lib/vmpooler/providers/vsphere.rb
```diff
@@ -839,7 +839,9 @@ def find_least_used_host(cluster, connection, datacentername)
         def find_cluster(cluster, connection, datacentername)
           datacenter = connection.serviceInstance.find_datacenter(datacentername)
           raise("Datacenter #{datacentername} does not exist") if datacenter.nil?
-          datacenter.hostFolder.children.find { |cluster_object| cluster_object.name == cluster }
+          found_cluster = datacenter.hostFolder.traverse(cluster)
```
sbeaulie: this requires changes to `cluster` from a named string to a `path` (see equivalent method in rbvmomi https://www.rubydoc.info/github/vmware/rbvmomi/RbVmomi%2FVIM%2FDatacenter:find_compute_resource). So other code changes up the stack would be needed.

My understanding is that it would return the first found item, not necessarily the right item, for example if you have multiple clusters in the datacenter. This is represented in the #find_cluster spec tests, so the PR here should not break those tests (or require changes in line with the logic change) 

https://github.com/puppetlabs/vmpooler/blob/98a84ab239c204da8022a87e0aa66307dd14f718/spec/unit/providers/vsphere_spec.rb#L2883

That being said I agree with the original intention of the PR to fix finding folders that have the same name as the target cluster, but I believe the solution should be with checking the type of the object found until it is a `RbVmomi::VIM::ComputeResource`

kevpl:
last comment is still current

kevpl:
last comment is still current. if this PR hasn't moved in a week, we'll close it due to inactivity.

kevpl:
closing due to inactivity. Once some movement has been made for this work, please re-open this PR or create a new one and we'll be happy to take a look at it.

PR #320, opened 2019-04-05
(POOLER-140) Ensure a running VM stays in a queue

mattkirby:
This commit updates how a VM is checked out to ensure that there is no window where the VM could be considered discovered, and therefore destroyed. Without this change the VM is retrieved by calling 'spop' on the ready queue, and then adding it to the running queue. This change moves to selecting the VM by retrieving the last member of the set, and moving it with 'smove' from ready to running. As a result of this change vmpooler moves from retrieving the VMs from the ready state randomly, to instead retrieve the oldest VM in the queue. This change should reduce churn where it would otherwise not be required to satisfy demand.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -68,9 +68,9 @@ module Vmpooler
       end
 
       template_backends.each do |template_backend|
-        vm = backend.spop("vmpooler__ready__#{template_backend}")
+        vm = backend.smembers("vmpooler__ready__#{template_backend}")[-1]
         if vm
-          backend.sadd("vmpooler__running__#{template_backend}", vm)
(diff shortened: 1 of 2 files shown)


lib/vmpooler/api/v1.rb
```diff
@@ -68,9 +68,9 @@ def fetch_single_vm(template)
       end
 
       template_backends.each do |template_backend|
-        vm = backend.spop("vmpooler__ready__#{template_backend}")
+        vm = backend.smembers("vmpooler__ready__#{template_backend}")[-1]
```
sbeaulie: is `smember()` nil-safe?

mattkirby: Just tested, and it looks like it should be alright. When smembers is empty it returns an empty array, and requesting the last element of that empty array produces nil, so I think it should be ok.

PR #329, opened 2019-07-16
(POOLER-140) Ensure a VM is alive at checkout

mattkirby:
This commit duplicates the vm_ready? check to the API layer to allow for API to validate that a VM is alive at checkout. Without this change API relies upon the checks in pool_manager validating pools. This change should allow for additional insight into whether a machine is in a ready state and resopnding at checkout time.

### DIFF
--- a/spec/integration/api/v1/vm_spec.rb
+++ b/spec/integration/api/v1/vm_spec.rb
@@ -15,7 +15,7 @@ describe Vmpooler::API::V1 do
       {
         config: {
           'site_name' => 'test pooler',
-          'vm_lifetime_auth' => 2,
+          'vm_lifetime_auth' => 2
         },
         pools: [
           {'name' => 'pool1', 'size' => 5},
@@ -28,6 +28,7 @@ describe Vmpooler::API::V1 do
       }
     }
     let(:current_time) { Time.now }
+    let(:vmname) { 'abcdefghijkl' }
 
     before(:each) do
       app.settings.set :config, config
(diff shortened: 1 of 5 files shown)


mattkirby:
@kevpl I found that when the allow statement into a before(:each) block that it caused additional failures. I think this was because cases that were supposed to fail would show vm_ready? as true. I think it would be possible to reduce duplication by restructuring the tests to confine the before statements to where the addition is needed. It seemed like more work than what I ended up with though, so unless your preference is strong I think specifying the addition for each test works ok. Feel free to point out if I'm overcomplicating it though.

spec/integration/api/v1/vm_spec.rb
```diff
@@ -56,32 +57,37 @@ def app()
     end
 
     describe 'POST /vm' do
+
+      let(:socket) { double('socket') }
       it 'returns a single VM' do
-        create_ready_vm 'pool1', 'abcdefghijklmnop'
+        create_ready_vm 'pool1', vmname
+        allow_any_instance_of(Vmpooler::API::Helpers).to receive(:open_socket).and_return(socket)
```
mattkirby: I could not find a way to use allow to stub the return value of the open_socket method without using `allow_any_instance_of`. I suspect there's something I don't understand with where sinatra is loading the helpers, and that if I were to understand that better there is likely another way to get at this. I'm all for a cleaner solution, but after several days of banging my head against this I did not find one.

PR #333, opened 2019-08-06
Make it possible to disable linked clones

mattkirby:
This commit adds a new configuration parameter to allow setting whether to create linked clones on a global, or per pool basis. Without this change vmpooler would always attempt to create linked clones. The default behavior of creating linked clones is preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,15 @@ The format is based on
 Tracking in this Changelog began for this project with the tagging of version 0.1.0.
 If you're looking for changes from before this, refer to the project's
 git logs & PR history.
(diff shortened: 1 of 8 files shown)


mattkirby:
@sbeaulie it's possible, but we would have to bubble up the disk backend preference from clone_vm in vsphere to pool_manager, which is what is creating the logs. I'm not sure it's worth it.
They're not even close in terms of the time it takes. A linked clone may take 5-10 seconds to clone. A full clone will take 90-180 seconds. Additionally, we can get the running configuration to see if linked clones are enabled for a pool in question.
What do you think?

sbeaulie:
@mattkirby as discussed this is not a show stopper, it is a nice to have. Since its not trivial to make this feature, I'm okay to go ahead with this change.

PR #338, opened 2019-08-30
(POOLER-148) Fix undefined variable bug in _check_ready_vm.

quorten:
The host['boottime'] variable in the function _check_ready_vm no longer
has its parent object in reference due to the refactoring in pull
request #269.  So in order to get the same information without the
performance impact from duplicate object lookups, we get similar
information from the time that the VM is ready.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -156,8 +156,14 @@ module Vmpooler
         $redis.hset('vmpooler__vm__' + vm, 'check', Time.now)
         # Check if the hosts TTL has expired
         if ttl > 0
-          # host['boottime'] may be nil if host is not powered on
-          if ((Time.now - host['boottime']) / 60).to_s[/^\d+\.\d{1}/].to_f > ttl
+          # 'boottime' may be nil if host is not powered on
+          boottime = $redis.hget("vmpooler__vm__#{vm}", 'ready')
+          if boottime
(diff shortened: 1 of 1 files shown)


quorten:
Hello, I'm new to contributing to PuppetLabs vmpooler and this is my first pull request.  I'm not entirely familiar with the PuppetLabs JIRA setup, so I put together the JIRA ticket the best I could and hopefully we can work from there.

lib/vmpooler/pool_manager.rb
```diff
@@ -156,8 +156,14 @@ def _check_ready_vm(vm, pool_name, ttl, provider)
         $redis.hset('vmpooler__vm__' + vm, 'check', Time.now)
         # Check if the hosts TTL has expired
         if ttl > 0
-          # host['boottime'] may be nil if host is not powered on
-          if ((Time.now - host['boottime']) / 60).to_s[/^\d+\.\d{1}/].to_f > ttl
+          # 'boottime' may be nil if host is not powered on
```
sbeaulie: Thanks for your contribution @quorten and for squishing this bug! Could we update the comment to be more relevant?
```suggestion
          # if 'boottime' is nil, set bootime to beginning of unix epoch, forces TTL to be assumed expired
```

quorten: Sure thing, looks good.

PR #341, opened 2019-10-23
(QENG-7530) Make VM names more human readable

highb:
Prior to this commit hostnames for VMs provisioned by vmpooler were 15
random characters. This is difficult for humans to tell apart.

This commit updates the naming to use the `spicy-proton` gem to generate
adjective noun pair names for the VMs, which I think would be easier for
humans to tell apart, as well as fun and memorable to say.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -1,4 +1,5 @@
 require 'vmpooler/providers'
+require 'spicy-proton'
 
 module Vmpooler
   class PoolManager
@@ -28,6 +29,9 @@ module Vmpooler
       @reconfigure_pool = {}
 
       @vm_mutex = {}
+
+      # Name generator for generating host names
+      @name_generator = Spicy::Proton.new
 
       # load specified providers from config file
       load_used_providers
@@ -265,8 +269,8 @@ module Vmpooler
 
     def _clone_vm(pool_name, provider)
       # Generate a randomized hostname
-      o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
-      new_vmname = $config[:config]['prefix'] + o[rand(25)] + (0...14).map { o[rand(o.length)] }.join
+      random_name = [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
+      new_vmname = $config[:config]['prefix'] + random_name
 
       # Add VM to Redis inventory ('pending' pool)
       $redis.sadd('vmpooler__pending__' + pool_name, new_vmname)
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'net-ldap', '~> 0.16'
 gem 'statsd-ruby', '~> 1.4.0', :require => 'statsd'
 gem 'connection_pool', '~> 2.2'
 gem 'nokogiri', '~> 1.8'
+gem 'spicy-proton', '2.1.1'
 
 group :development do
   gem 'pry'


highb:
Some example pairs (a preview of the fun to come):
```
[3] pry(main)> @name_generator = Spicy::Proton.new
=> #<Spicy::Proton:0x00007fec88ea8ce8 @corpus=#<Spicy::Memory::Corpus:0x00007fec88ea8cc0 @lists={}>>
[4] pry(main)> random_name = @name_generator.format('%a-%n')
=> "nutrient-sera"
[5] pry(main)> random_name = @name_generator.format('%a-%n')
=> "counterfeit-badmen"
[6] pry(main)> random_name = @name_generator.format('%a-%n')
=> "faulknerian-bebop"
[7] pry(main)> random_name = @name_generator.format('%a-%n')
=> "uncertified-trumpet"
[8] pry(main)> random_name = @name_generator.format('%a-%n')
=> "photogenic-repentance"
[9] pry(main)> random_name = @name_generator.format('%a-%n')
=> "chunky-handkerchief"
[10] pry(main)> random_name = @name_generator.format('%a-%n')
=> "practical-sovereign"
```

highb:
I've restricted the maximum word lengths so that the hostname shouldn't exceed 15 characters. Here's some example names:
```
[2] pry(main)> @name_generator = Spicy::Proton.new
=> #<Spicy::Proton:0x00007fd050576a20 @corpus=#<Spicy::Memory::Corpus:0x00007fd0505769f8 @lists={}>>
[3] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "deluxe-mantel"
[4] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "wise-showing"
[5] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "rocky-lien"
[6] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "tragic-bundle"
[7] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "genial-comb"
[8] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "forlorn-vasa"
[9] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "south-amp"
[10] pry(main)> [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
=> "bestial-finding"
```

lib/vmpooler/pool_manager.rb
```diff
@@ -265,8 +269,8 @@ def clone_vm(pool_name, provider)
 
     def _clone_vm(pool_name, provider)
       # Generate a randomized hostname
-      o = [('a'..'z'), ('0'..'9')].map(&:to_a).flatten
-      new_vmname = $config[:config]['prefix'] + o[rand(25)] + (0...14).map { o[rand(o.length)] }.join
+      random_name = [@name_generator.adjective(max: 7), @name_generator.noun(max: 7)].join('-')
+      new_vmname = $config[:config]['prefix'] + random_name
```
genebean: @highb This line would allow the name to exceed 15 characters... how about we modify this like so:

```ruby
netbios_limit = 15
adj_len = (netbios_limit - $config[:config]['prefix'].length) / 2
random_name = [@name_generator.adjective(max: adj_len), @name_generator.noun(max: adj_len)].join('-')
new_vmname = $config[:config]['prefix'] + random_name
```

highb:
```
➜  docker git:(imp/master/QENG-7530_human_readable_names) ping ionic-bank.delivery.puppetlabs.net                     ionic-bank.delivery.puppetlabs.net
PING ionic-bank.delivery.puppetlabs.net (10.16.119.101): 56 data bytes
64 bytes from 10.16.119.101: icmp_seq=0 ttl=58 time=1.189 ms
64 bytes from 10.16.119.101: icmp_seq=1 ttl=58 time=1.069 ms
^C
--- ionic-bank.delivery.puppetlabs.net ping statistics ---
2 packets transmitted, 2 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 1.069/1.129/1.189/0.060 ms
➜  docker git:(imp/master/QENG-7530_human_readable_names) ssh ionic-bank.delivery.puppetlabs.net                      ionic-bank.delivery.puppetlabs.net
Last login: Thu Oct 24 21:31:35 2019 from 10.0.17.1
[root@ionic-bank ~]#
```
Booted up the docker compose vmpooler with the eso password configured and it appears to provision fine. 🎉

lib/vmpooler/pool_manager.rb
highb: This doesn't seem to be a problem with the current 15 character random name. What does NetBIOS do if the name exceeds 15 characters? Truncate it? Do we actually use prefixes on any pools?

highb: I asked/looked around and we don't use the prefix. The prefix config value was added by a community member so I assume they don't mind if the hostname exceeds 15 characters.
Here's the relevant commit: https://github.com/puppetlabs/vmpooler/commit/10e507c262ae496c88b31e5c03d126ecc51addf0

genebean: We have customers running this who I know for a fact use Windows machines so I thought it best to make sure that if we were aiming for a 15 character or less name that that was enforced even if those users add a prefix.

highb: The problem with the `(netbios_limit - $config[:config]['prefix'].length) / 2` formula is that _I think_ it is going to alternate between being 15 and 16 characters depending on if the prefix is odd or even due to integer rounding and the hypen. See this little pry experiment:
```
[10] pry(main)> prefix_len = 0; (((15 - prefix_len) / 2) * 2) + 1 + prefix_len
=> 15
[11] pry(main)> prefix_len = 1; (((15 - prefix_len) / 2) * 2) + 1 + prefix_len
=> 16
[12] pry(main)> prefix_len = 2; (((15 - prefix_len) / 2) * 2) + 1 + prefix_len
=> 15
[13] pry(main)> prefix_len = 3; (((15 - prefix_len) / 2) * 2) + 1 + prefix_len
=> 16
[14] pry(main)> prefix_len = 4; (((15 - prefix_len) / 2) * 2) + 1 + prefix_len
=> 15
[15] pry(main)> prefix_len = 5; (((15 - prefix_len) / 2) * 2) + 1 + prefix_len
=> 16
```
So we'd need a different formula there. I'm not sure if we should put in the effort of adding fiddly logic like this if users seem to be fine with the current prefix plus 15 character length.

genebean: It looks like changing to `adj_len = ((netbios_limit - prefix.length) / 2) - 1` fixes that @highb. I tested with this:

```ruby
def new_name_info (prefix)
  netbios_limit = 15
  adj_len = ((netbios_limit - prefix.length) / 2) - 1
  # random_name = [@name_generator.adjective(max: adj_len), @name_generator.noun(max: adj_len)].join('-')
  random_name = ['a' * adj_len, 'a' * adj_len].join('-')
  new_vmname = prefix + random_name
  puts "With a #{prefix.length}-charcater prefix, the new vmname is #{new_vmname} and is #{new_vmname.length} characters long"
end

new_name_info('')
new_name_info('b')
new_name_info('bb')
new_name_info('bbb')
new_name_info('bbbb')
new_name_info('bbbbb')
```

```
╔ ☕️  gene:~
╚ᐅ ruby ~/Downloads/test.rb
With a 0-charcater prefix, the new vmname is aaaaaa-aaaaaa and is 13 characters long
With a 1-charcater prefix, the new vmname is baaaaaa-aaaaaa and is 14 characters long
With a 2-charcater prefix, the new vmname is bbaaaaa-aaaaa and is 13 characters long
With a 3-charcater prefix, the new vmname is bbbaaaaa-aaaaa and is 14 characters long
With a 4-charcater prefix, the new vmname is bbbbaaaa-aaaa and is 13 characters long
With a 5-charcater prefix, the new vmname is bbbbbaaaa-aaaa and is 14 characters long
```

highb: That has the same problem, and reduces the corpus of random words by a large factor. I really don't think this is worth worrying about until someone that actually uses the prefix config setting opens an issue explaining their use case.

sbeaulie: Feels like a spec would work well here

One edge case is if the prefix is large I'm not sure what the generator does with maxes set to 0, 1 or 2

While we are at it we could also adjust the noun's max length after we know the size of the adjective eg if the adjective is 3 chars then the the noun can be 11 char max?

PR #342, opened 2019-10-25
Add spicy-proton to vmpooler.gemspec

mattkirby:
This commit adds the spicy-proton gem to vmpooler.gemspec. Without this change the spicy-proton gem is in the Gemfile, but not the gemspec, causing issues when deploying from gem.


mattkirby:
No worries, I reviewed it and did not notice until I hit the issue.

Issue #205, opened 2017-03-14
find_cluster in vsphere_helper doesn't support host folders


glennsarti:
It is possible to create host folders so that clusters and single hosts are in a folder hierarchy.

Currently the `find_cluster` function only looks in the root folder, not sub folders
https://github.com/puppetlabs/vmpooler/blob/master/lib/vmpooler/vsphere_helper.rb#L244-L247

While this is not a common deployment scenario, it is possible that clusters are being missed in this scenario

PR #348, opened 2019-11-26
Support nested host folders in find_cluster()

seanmil:
Search the root and any subfolders for cluster or host resources.

This addresses issue #205


sbeaulie:
@mattkirby if you could review this and comment on if it would work on our infra, that would be great!

PR #349, opened 2019-12-02
(POOLER-123) Implement a max TTL

sbeaulie:
Before this change, we could checkout a vm and set the lifetime to a
very high number which would esssentially keep the vm running forever.
Now implementing a config setting max_lifetime_upper_limit which enforces
a maximum lifetime in hours both for initial checkout and extending a
running vm

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -22,6 +22,8 @@ services:
       - redislocal
   redislocal:
     image: redis
+    # Uncomment this if you don't want the redis data to persist
+    #command: "redis-server --save '' --appendonly no"
     ports:
       - "6379:6379"
     networks:
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -875,7 +875,7 @@ module Vmpooler
       status 404
       result = { 'ok' => false }
 
-      failure = false
+      failure = []
 
       params[:hostname] = hostname_shorten(params[:hostname], config['domain'])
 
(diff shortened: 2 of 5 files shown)


highb:
Hmmm... seems that when I spin this up with `max_lifetime_upper_limit: 9000` set in the config I can no longer change the lifetime at all? Here's how it works normally against prod:
```
➜  vmpooler git:(POOLER-123) floaty get centos-7-64 --service vm
- diffuse-pizza.delivery.puppetlabs.net (centos-7-64)
➜  vmpooler git:(POOLER-123) floaty modify diffuse-pizza.delivery.puppetlabs.net --service vm --lifetime 10

Successfully modified VM diffuse-pizza.delivery.puppetlabs.net.
Use `floaty list --active` to see the results.
➜  vmpooler git:(POOLER-123) floaty list --active --service vm

Your VMs on vmpooler.delivery.puppetlabs.net:
- diffuse-pizza.delivery.puppetlabs.net (centos-7-x86_64, 0.01/10 hours)
```

And here's with the changes running locally:
```
➜  vmpooler git:(POOLER-123) floaty list --active --service vmlocal
Your VMs on localhost:
- amateur-voyage.delivery.puppetlabs.net (centos-7-x86_64, 0.08/2 hours)
➜  vmpooler git:(POOLER-123) floaty modify amateur-voyage.delivery.puppetlabs.net --lifetime 8999 --trace

Successfully modified VM amateur-voyage.delivery.puppetlabs.net.
Use `floaty list --active` to see the results.
➜  vmpooler git:(POOLER-123) floaty list --active --service vmlocal
Your VMs on localhost:
- amateur-voyage.delivery.puppetlabs.net (centos-7-x86_64, 0.09/2 hours)
```

sbeaulie:
I should check what api command floaty uses? Did you put `max_lifetime_upper_limit` as a child of the `config:` key, and not top scope?

highb:
Like this?
```
---
:config:
  site_name: '<b>vmpooler-provisioner-highb</b>.delivery.puppetlabs.net'
  domain: 'delivery.puppetlabs.net'
  logfile: '/var/log/vmpooler.log'
  clone_target: 'acceptance4'
  task_limit: 10
  vm_lifetime: 1
  vm_lifetime_auth: 2
  vm_checktime: 1
  max_tries: 3
  retry_factor: 10
  create_folders: true
  backend_weight:
    'acceptance1': 5
    'acceptance2': 6
    'acceptance3': 6
  purge_unconfigured_folders: true
  experimental_features: true
  migration_limit: 10
  max_lifetime_upper_limit: 9000
```

highb:
As for floaty, I think it just `PUTS` to the `vm` endpoint https://github.com/briancain/vmfloaty/blob/master/lib/vmfloaty/pooler.rb#L54
```
  def self.modify(verbose, url, hostname, token, modify_hash)
    raise TokenError, 'Token provided was nil. Request cannot be made to modify vm' if token.nil?

    modify_hash.keys.each do |key|
      raise ModifyError, "Configured service type does not support modification of #{key}." unless %i[tags lifetime disk].include? key
    end
...
  response = conn.put do |req|
      req.url "vm/#{hostname}"
      req.body = modify_hash.to_json
    end
```

lib/vmpooler/api/v1.rb
```diff
@@ -892,6 +892,22 @@ def invalid_pool(payload)
             when 'lifetime'
               need_token! if Vmpooler::API.settings.config[:auth]
 
+              # in hours, defaults to one week
+              max_lifetime_upper_limit = config[:max_lifetime_upper_limit]
```
highb: Ah, I found the issue. This key is a string, not a symbol. When I switched that out, it works.

highb:
I've switched out the symbol for a string in the config hash reference and added a little better error messaging so a user knows what actually went wrong. Also, vmfloaty didn't correctly handle when the PUT came back with `ok: false` so I've put up a PR to fix that up a bit. https://github.com/briancain/vmfloaty/pull/55

docker/docker-compose.yml
```diff
@@ -22,6 +22,8 @@ services:
       - redislocal
   redislocal:
     image: redis
+    # Uncomment this if you don't want the redis data to persist
+    #command: "redis-server --save '' --appendonly no"
```
highb: Oh and this little line was helpful for me to ensure that redis wasn't persisting between testing runs, as it was getting a strange no space left on device error. My computer has plenty of room, so I'm not sure where that is coming from all of a sudden.

lib/vmpooler/api/v1.rb
sbeaulie: good catch. That was an issue with testing this as well, since we set the parameters and I wasn't sure if it was a symbol or a string. Thanks!

docker/docker-compose.yml
sbeaulie: check the docker settings for how much space is allowed for all containers? I had to bump mine up a few times.

sbeaulie:
Thanks for the commits @highb 

high five

docker/docker-compose.yml
highb: Ah, I didn't even know that was a setting. Thanks!

PR #352, opened 2020-01-31
Add vmpooler flush script

highb:


mattkirby:
@highb would it be better to be able to request this via the API? I was thinking we can reuse what we do when we swap out a template, as an example, and allow calling that for a pool.

highb:
I'm sure it would be. I just wanted to capture the one-liner you posted to the channel so we at least have a record of that somewhere in case we need to use it in a pinch 😄 . In the long run, we should definitely make an API endpoint. Because yeah I agree, I think we could just call the `flush` method that is already used in several places.

mattkirby:
I put up a PR to offer a mechanism for this via API, #353 . @highb if that change is accepted do you think that we should add this script as well?

highb:
Hm... I don't think there's much value to a script if we have a document API endpoint. I'll work on reviewing the endpoint.

highb:
Closing this as unnecessary now that there is an API endpoint #353

PR #353, opened 2020-02-12
(POOLER-153) Add endpoint for resetting a pool

mattkirby:
This commit adds a capability to vmpooler to reset a pool, deleting its ready and pending instances and replacing them with fresh ones. Without this change vmpooler does not offer a mechanism to reset a pool without also changing its template.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -773,3 +773,29 @@ $ curl -X POST -H "Content-Type: application/json" -d '{"debian-7-i386":"templat
   "ok": true
 }
 ```
+
+##### POST /poolreset
+
+Clear all pending and ready instances in a pool, and deploy replacements
+
+All pool reset requests must be for pools that exist in the vmpooler configuration running, or a 404 code will be returned.
+
+When a pool reset is requested a 201 status will be returned.
+
+A pool reset will cause vmpooler manager to log that it has cleared ready and pending instances.
+
+For poolreset to be available it is necessary to enable experimental features. Additionally, the request must be performed with an authentication token when authentication is configured.
+
+Responses:
+* 201 - Pool reset requested received
+* 400 - An invalid configuration was provided causing requested changes to fail
+* 404 - An unknown error occurred
+* 405 - The endpoint is disabled because experimental features are disabled
+```
+$ curl -X POST -H "Content-Type: application/json" -d '{"debian-7-i386":"1"}' --url https://vmpooler.example.com/api/v1/poolreset
+```
+```json
+{
+  "ok": true
+}
+```
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -200,6 +200,17 @@ module Vmpooler
       result
     end
 
+    def reset_pool(payload)
+      result = { 'ok' => false }
+
+      payload.each do |poolname, count|
+        backend.sadd('vmpooler__poolreset', poolname)
+      end
+      status 201
+      result['ok'] = true
+      result
+    end
+
     def update_clone_target(payload)
       result = { 'ok' => false }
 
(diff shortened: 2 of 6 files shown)


docs/API.md
```diff
@@ -773,3 +773,29 @@ $ curl -X POST -H "Content-Type: application/json" -d '{"debian-7-i386":"templat
...
+
+For poolreset to be available it is necessary to enable experimental features. Additionally, the request must be performed with an authentication token when authentication is configured.
+
+Responses:
+* 201 - Pool reset requested received
+* 400 - An invalid configuration was provided causing requested changes to fail
+* 404 - An unknown error occurred
+* 405 - The endpoint is disabled because experimental features are disabled
+```
+$ curl -X POST -H "Content-Type: application/json" -d '{"debian-7-i386":"1"}' --url https://vmpooler.example.com/api/v1/poolreset
```
sbeaulie: I see that the number (poolcount) is not used, but is it needed to pass the validation? would a JSON like `{"debian-7-i386"}` also work?

lib/vmpooler/api/v1.rb
```diff
@@ -200,6 +200,21 @@ def update_pool_template(payload)
...
 
+    def reset_pool(payload)
+      result = { 'ok' => false }
+
+      pool_index = pool_index(pools)
+      pools_reset = 0
+
+      payload.each do |poolname, count|
+        backend.sadd('vmpooler__poolreset', poolname)
+        pools_reset += 1
```
sbeaulie: this `pools_reset` is not used

lib/vmpooler/api/v1.rb
```diff
@@ -200,6 +200,21 @@ def update_pool_template(payload)
...
+        status 201
```
sbeaulie: status could be outside the loop

lib/vmpooler/api/v1.rb
sbeaulie: maybe increase a metric instead?

docs/API.md
mattkirby: @sbeaulie, I would have preferred to skip the number, but yes, it being json made me thing that is necessary. I did spend some time attempting to replicate the behavior of the vm endpoint where we can pass a pool title as a part of the URL, instead of having to pass in a data block. However, I was not able to make this work, so eventually accepted having the number so I could push this over the line.

lib/vmpooler/api/v1.rb
mattkirby: Re-reading this it looks like we can just remove it. With poolsize and template it is possible a change is not made depending on the data you supply. With poolreset, it always resets the pool upon request. So, how many of the requested pools were reset is a moot point.

lib/vmpooler/api/v1.rb
mattkirby: Good point. I moved it out.

lib/vmpooler/api/v1.rb
```diff
@@ -1063,6 +1074,37 @@ def invalid_pool(payload)
...
     end
 
+    post "#{api_prefix}/poolreset/?" do
+      content_type :json
+      result = { 'ok' => false }
+
+      if config['experimental_features']
+        need_token! if Vmpooler::API.settings.config[:auth]
+
+        payload = JSON.parse(request.body.read)
```
highb: Since this payload isn't `rescue`d, a user will get a mysterious 500 error if they post bad JSON.

mattkirby: It looks like this is something that we do throughout vmpooler and will affect other POST methods in the vmpooler API. Should we try to rescue this and give a 500 error instead?

highb: Rescue it, I'd set a `400` (Bad request) with a message like `JSON payload could not be parsed`. Just something easy so that an API user knows there is something they can do to fix their request and that the server isn't just angry. 😄 

As for the other API endpoints, ideally I'd like if all of them had better error handling and user-friendly error messages but I figure we should provide some better messaging as we add new endpoints.

mattkirby: Thanks. I have updated this so poolreset will give a 400 and a message if bad JSON is provided. I tested it and validated I get the 400 and message about JSON not being able to be parsed when bad JSON is provided.

lib/vmpooler/api/v1.rb
```diff
@@ -1063,6 +1074,45 @@ def invalid_pool(payload)
...
+
+        begin
+          payload = JSON.parse(request.body.read)
+        rescue JSON::ParserError
+          status 400
+          result = {
+            'ok' => false,
+            'message' => 'JSON payload could not be parsed'
+          }
+        end
```
highb: Since the `begin`/`rescue` block ends here, won't the status always ends up getting overwritten with `404`, since `payload` will be `nil`? Perhaps the `begin` block should also include the `payload` handling?

mattkirby: Good point! Another change incoming.

mattkirby:
I think this is finally ready to go. Please let me know if I've missed anything else.

PR #356, opened 2020-03-04
Release 0.10.3

highb:
Updates for release 0.10.3, tagging will be handled by the job.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,7 +17,7 @@ Style/Documentation:
   Enabled: false
 
 # Line length is not useful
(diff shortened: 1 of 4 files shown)


highb:
Rubocop is completely broken due to some cop name changes, going to fix that and maybe run auto-corrects then update the changelog.

highb:
There are way too many issues to auto-correct and some of them are pretty significant code changes that we'd want to test out more extensively, so I'll delay doing an autocorrect for another release so that we can have the migration fix in the meanwhile.

PR #357, opened 2020-03-05
(POOLER-157) Add extra_config option to vmpooler

mattkirby:
This commit adds the extra_config option to vmpooler to allow specifying additional configuration files to load from. Without this change vmpooler does not offer a mechanism to provide additional configuration files for the application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,19 +13,19 @@ git logs & PR history.
 
 ## [Unreleased](https://github.com/puppetlabs/vmpooler/compare/0.10.3...master)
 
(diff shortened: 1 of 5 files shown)


mattkirby:
I adjusted the changelog so it jumps from 0.9.1 to 0.10.2 so it is easier to follow the feature changes. Please let me know if this change seems unreasonable and I can remove it.

highb:
@mattkirby Heh, honestly I wasn't sure if making an extra note for the versions with no changes was too verbose. I think being succinct like this is probably a better call.

PR #359, opened 2020-03-05
Fix Rubocop "safe" auto-corrections

highb:
Generated using `bundle exec rubocop --safe --auto-correct`

Oh and:
```20 files inspected, 427 offenses detected, 344 offenses corrected```

PR #360, opened 2020-03-05
"Unsafe" rubocop fixes

highb:
Adds the remaining "unsafe" fixes that aren't included in #359

On top of the changes from #359:
```20 files inspected, 151 offenses detected, 46 offenses corrected```

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Vmpooler
   class API
   class V1 < Sinatra::Base
@@ -936,7 +938,7 @@ module Vmpooler
           end
         end
 
-        if failure.size > 0
+        if !failure.empty?
           status 400
           result['failure'] = failure
         else
--- a/lib/vmpooler/providers.rb
+++ b/lib/vmpooler/providers.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'pathname'
 
 module Vmpooler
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

File: lib/vmpooler/providers.rb [hunk 1]
```diff
@@ -38,7 +40,7 @@ def load_from_gems(name = nil)
         # we don't exactly know if the provider name matches the main file name that should be loaded
         # so we use globs to get everything like the name
         # this could mean that vsphere5 and vsphere6 are loaded when only vsphere5 is used
-        Dir.glob(File.join(gem_path, "*#{name}*.rb")).each do |file|
+        Dir.glob(File.join(gem_path, "*#{name}*.rb")).sort.each do |file|
```
highb: This one is interesting:
```
lib/vmpooler/providers.rb:40:9: W: [Corrected] Lint/NonDeterministicRequireOrder: Sort files before requiring them.
        Dir.glob(File.join(gem_path, "*#{name}*.rb")).each do |file|
```

mattkirby: I'd imagine it's so load order is consistent. It would probably make it harder to troubleshoot if the load order was inconsistent and a gem were part of the problem.

lib/vmpooler/api/v1.rb
```diff
@@ -936,7 +938,7 @@ def invalid_pool(payload)
           end
         end
 
-        if failure.size > 0
+        if !failure.empty?
```
mattkirby: Why does this become !failure? I'm confused here.

PR #361, opened 2020-03-06
Manual Rubocop Fixes

highb:
This should fix the remaining Rubocop issues that couldn't be autocorrected.

Still working on the last couple of issues, I'll remove the DNM tag when those are finished.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -330,11 +330,11 @@ module Vmpooler
         $logger.log('s', "[+] [#{pool_name}] '#{new_vmname}' cloned in #{finish} seconds")
 
         $metrics.timing("clone.#{pool_name}", finish)
-      rescue StandardError => _e
+      rescue StandardError
         $redis.srem("vmpooler__pending__#{pool_name}", new_vmname)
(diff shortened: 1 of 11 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -369,7 +369,7 @@ def _destroy_vm(vm, pool, provider)
 
         $redis.srem('vmpooler__completed__' + pool, vm)
 
-        finish = format('%.2f', Time.now - start)
+        finish = format('%<time>.2f', time: Time.now - start)
```
mattkirby: Presumably this was to resolve a failure. Does the addition of 'time:' change the behavior at all?

highb: Yep, the [StyleFormatStringToken](https://rubocop.readthedocs.io/en/latest/cops_style/#styleformatstringtoken) cop prefers named parameters and it shouldn't change the behavior at all. I can verify that the output looks the same now.

highb: If you prefer not having named parameters in the format strings, we could also do `EnforcedStyle: unannotated`

mattkirby: I'm fine with it, I just wanted to make sure that I followed.

PR #363, opened 2020-03-12
Add a .dockerignore file

mattkirby:


.dockerignore
```diff
@@ -0,0 +1,13 @@
+**/*.yml
+**/*.yaml
+**/*.md
+**/*example
+Dockerfile?
```
highb: Looks like the dockerfiles are still getting picked up:
```
root@304bbf971d5d:/# ls docker/
Dockerfile  Dockerfile-aio  Dockerfile_local  docker-entrypoint.sh
```
I suspect that if you wanted to ignore everywhere you'd need `**/Dockerfile*`

mattkirby: Good catch. I updated it.

PR #364, opened 2020-03-12
Add support for ldaps on port 636

mattkirby:
This commit adds support for ldaps on port 636. Without this change vmpooler assumes the auth method is start_tls, which does not work with ldaps.

### DIFF
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -57,11 +57,13 @@ module Vmpooler
       end
 
       def authenticate_ldap(port, host, user_object, base, username_str, password_str)
+        auth_method = :start_tls if port == 389
+        auth_method = :simple_tls if port == 636
         ldap = Net::LDAP.new(
           :host => host,
           :port => port,
           :encryption => {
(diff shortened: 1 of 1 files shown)


lib/vmpooler/api/helpers.rb
```diff
@@ -57,11 +57,13 @@ def authorized?
       end
 
       def authenticate_ldap(port, host, user_object, base, username_str, password_str)
+        auth_method = :start_tls if port == 389
+        auth_method = :simple_tls if port == 636
```
highb: It looks like `sasl` is another valid method. Would we want to allow a user override via the config for this or is `simple_tls` good enough for now?

mattkirby: Yes. There are probably several different options a user may want to provide. When using ldaps on 636 one concern is additional steps are required to validate that your server is who it should be. This is probably not enough. Also, my original reason for opening this (our LDAP server only being open on 636) was addressed, so I can use 389 for this instead. I'm going to close this for now instead of adding the additions that would make this support more well rounded. If this comes up we can revisit it again then.

PR #367, opened 2020-03-17
Add Dependabot to keep gems updated

genebean:


mattkirby:
I did not know about this. Thanks @genebean !

PR #368, opened 2020-03-17
Remove providers addition to docker-compose.yml

mattkirby:


mattkirby:
I accidentally added a few values that were in use for testing to the docker-compose.yml file intended to support local testing. These additions will impact testing for other users, so this commit removes the new values to ensure that my testing does not pollute the default values.

highb:
Ah, I also had the providers.yaml setup from testing the other PR so I didn't notice. Whoops!

PR #371, opened 2020-04-15
(POOLER-161) Fix extending vm lifetime when max lifetime is set

sbeaulie:
Before this PR, the current running time was being inspected to decide if the
vm lifetime could be extended. But since vm lifetime is absolute and not relative
this check is now removed.

### DIFF
--- a/spec/integration/api/v1/vm_hostname_spec.rb
+++ b/spec/integration/api/v1/vm_hostname_spec.rb
@@ -134,7 +134,7 @@ describe Vmpooler::API::V1 do
           expect(vm['lifetime']).to eq("20000")
         end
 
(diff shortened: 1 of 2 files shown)


spec/integration/api/v1/vm_hostname_spec.rb
```diff
@@ -134,7 +134,7 @@ def app()
           expect(vm['lifetime']).to eq("20000")
         end
 
-        it 'does not allow a lifetime to be initially past config 168' do
+        it 'does not allow a lifetime to be initially past config max_lifetime_upper_limit' do
```
jcoconnor, line 137: Would it be worth putting the actual numerical limit here?

highb:
👍 Still works with a VM that has been up for two hours.

PR #374, opened 2020-04-23
(DIO-608) vmpooler SUT handed out multiple times

sbeaulie:
Before this change if the smove returned false, we would continue handing out the VM
which presumably could still be in the 'ready' state. Upon 'delete' that ready VM would not be
picked up and return a 404 which is consistent with the behavior seen. Adding a metric to keep
track of the smove failures since this is not expected. I think some API logging would be good
to add in the future.

### DIFF
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -83,8 +83,13 @@ module Vmpooler
           vms.reverse.each do |vm|
             ready = vm_ready?(vm, config['domain'])
             if ready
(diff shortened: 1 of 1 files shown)


sbeaulie:
If this is the root cause of the isue it will also be fixed since it won't hand over VMs that it couldnt smove in redis.
There is currently no logger in the API, and we always look at the manager logs, but I think the API could use some logging library too,  not sure if we have a pattern for sinatra apps.

PR #375, opened 2020-05-07
(POOLER-158) Add support for ondemand provisioning

mattkirby:
This change adds support to vmpooler for provisioning on demand. Additionally, this change adds redis connection pooling to pool_manager. Without this change vmpooler does not support provisioning on demand. Additionally, without this change the redis usage in pool_manager is not thread safe.

Similar changes should be made to vmpooler API as well, but have not been included here to limit the scope of change in this PR somewhat.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - VMPOOLER_DEBUG=true # for use of dummy auth
       - VMPOOLER_CONFIG_FILE=/etc/vmpooler/vmpooler.yaml
       - REDIS_SERVER=redislocal
-      - LOGFILE=/dev/stdout
+      - LOGFILE=/dev/null
     image: vmpooler-local
     depends_on:
       - redislocal
--- a/lib/vmpooler/api/v1.rb
+++ b/lib/vmpooler/api/v1.rb
@@ -42,6 +42,68 @@ module Vmpooler
       Vmpooler::API.settings.checkoutlock
     end
 
+    def get_template_aliases(template)
+      result = []
+      aliases = Vmpooler::API.settings.config[:alias]
+      if aliases
+        result += aliases[template] if aliases[template].is_a?(Array)
+        template_backends << aliases[template] if aliases[template].is_a?(String)
+      end
+      result
+    end
+
+    def get_pool_weights(template_backends)
+      pool_index = pool_index(pools)
+      weighted_pools = {}
+      template_backends.each do |t|
+        next unless pool_index.key? t
+
+        index = pool_index[t]
+        clone_target = pools[index]['clone_target'] || config['clone_target']
+        next unless config.key?('backend_weight')
+
+        weight = config['backend_weight'][clone_target]
+        if weight
+          weighted_pools[t] = weight
+        end
+      end
+      weighted_pools
+    end
+
+    def count_selection(selection)
+      result = {}
+      selection.uniq.each do |poolname|
+        result[poolname] = selection.count(poolname)
+      end
+      result
+    end
+
+    def evaluate_template_aliases(template, count)
+      template_backends = []
+      template_backends << template if backend.sismember('vmpooler__pools', template)
+      selection = []
+      aliases = get_template_aliases(template)
+      if aliases
+        template_backends += aliases
+        weighted_pools = get_pool_weights(template_backends)
+
+        pickup = Pickup.new(weighted_pools) if weighted_pools.count == template_backends.count
+        count.to_i.times do
+          if pickup
+            selection << pickup.pick
+          else
+            selection << template_backends.sample
+          end
+        end
+      else
+        count.to_i.times do
+          selection << template
+        end
+      end
+
+      count_selection(selection)
+    end
+
     def fetch_single_vm(template)
       template_backends = [template]
       aliases = Vmpooler::API.settings.config[:alias]
@@ -245,11 +307,9 @@ module Vmpooler
       pool_index = pool_index(pools)
       template_configs = backend.hgetall('vmpooler__config__template')
       template_configs&.each do |poolname, template|
-          if pool_index.include? poolname
-            unless pools[pool_index[poolname]]['template'] == template
-              pools[pool_index[poolname]]['template'] = template
-            end
-          end
+        next unless pool_index.include? poolname
+
+        pools[pool_index[poolname]]['template'] = template
       end
     end
 
@@ -257,11 +317,9 @@ module Vmpooler
       pool_index = pool_index(pools)
       poolsize_configs = backend.hgetall('vmpooler__config__poolsize')
       poolsize_configs&.each do |poolname, size|
-          if pool_index.include? poolname
-            unless pools[pool_index[poolname]]['size'] == size.to_i
-              pools[pool_index[poolname]]['size'] == size.to_i
-            end
-          end
+        next unless pool_index.include? poolname
+
+        pools[pool_index[poolname]]['size'] = size.to_i
       end
     end
 
@@ -269,12 +327,67 @@ module Vmpooler
       pool_index = pool_index(pools)
       clone_target_configs = backend.hgetall('vmpooler__config__clone_target')
       clone_target_configs&.each do |poolname, clone_target|
-          if pool_index.include? poolname
-            unless pools[pool_index[poolname]]['clone_target'] == clone_target
-              pools[pool_index[poolname]]['clone_target'] == clone_target
-            end
-          end
+        next unless pool_index.include? poolname
+
+        pools[pool_index[poolname]]['clone_target'] = clone_target
       end
+    end
+
+    def too_many_requested?(payload)
+      payload&.each do |_poolname, count|
+        next unless count.to_i > config['max_ondemand_instances_per_request']
+
+        return true
+      end
+      false
+    end
+
+    def generate_ondemand_request(payload)
+      result = { 'ok': false }
+
+      requested_instances = payload.reject { |k, _v| k == 'request_id' }
+      if too_many_requested?(requested_instances)
+        result['message'] = "requested amount of instances exceeds the maximum #{config['max_ondemand_instances_per_request']}"
+        status 403
+        return result
+      end
+
+      score = Time.now.to_i
+      request_id = payload['request_id']
+      request_id ||= generate_request_id
+      result['request_id'] = request_id
+
+      if backend.exists("vmpooler__odrequest__#{request_id}")
+        result['message'] = "request_id '#{request_id}' has already been created"
+        status 409
+        return result
+      end
+
+      status 201
+
+      platforms_with_aliases = []
+      requested_instances.each do |poolname, count|
+        selection = evaluate_template_aliases(poolname, count)
+        selection.map { |selected_pool, selected_pool_count| platforms_with_aliases << "#{poolname}:#{selected_pool}:#{selected_pool_count}" }
+      end
+      platforms_string = platforms_with_aliases.join(',')
+
+      return result unless backend.zadd('vmpooler__provisioning__request', score, request_id)
+
+      backend.hset("vmpooler__odrequest__#{request_id}", 'requested', platforms_string)
+      if Vmpooler::API.settings.config[:auth] and has_token?
+        backend.hset("vmpooler__odrequest__#{request_id}", 'token:token', request.env['HTTP_X_AUTH_TOKEN'])
+        backend.hset("vmpooler__odrequest__#{request_id}", 'token:user',
+                     backend.hget('vmpooler__token__' + request.env['HTTP_X_AUTH_TOKEN'], 'user'))
+      end
+
+      result['domain'] = config['domain'] if config['domain']
+      result[:ok] = true
+      result
+    end
+
+    def generate_request_id
+      SecureRandom.uuid
     end
 
     get '/' do
@@ -395,7 +508,7 @@ module Vmpooler
           end
 
           # for backwards compatibility, include separate "empty" stats in "status" block
-          if ready == 0
+          if ready == 0 && max != 0
             result[:status][:empty] ||= []
             result[:status][:empty].push(pool['name'])
 
@@ -689,6 +802,88 @@ module Vmpooler
       JSON.pretty_generate(result)
     end
 
+    post "#{api_prefix}/ondemandvm/?" do
+      content_type :json
+
+      need_token! if Vmpooler::API.settings.config[:auth]
+
+      result = { 'ok' => false }
+
+      begin
+        payload = JSON.parse(request.body.read)
+
+        if payload
+          invalid = invalid_templates(payload.reject { |k, _v| k == 'request_id' })
+          if invalid.empty?
+            result = generate_ondemand_request(payload)
+          else
+            result[:bad_templates] = invalid
+            invalid.each do |bad_template|
+              metrics.increment('ondemandrequest.invalid.' + bad_template)
+            end
+            status 404
+          end
+        else
+          metrics.increment('ondemandrequest.invalid.unknown')
+          status 404
+        end
+      rescue JSON::ParserError
+        status 400
+        result = {
+          'ok' => false,
+          'message' => 'JSON payload could not be parsed'
+        }
+      end
+
+      JSON.pretty_generate(result)
+    end
+
+    post "#{api_prefix}/ondemandvm/:template/?" do
+      content_type :json
+      result = { 'ok' => false }
+
+      need_token! if Vmpooler::API.settings.config[:auth]
+
+      payload = extract_templates_from_query_params(params[:template])
+
+      if payload
+        invalid = invalid_templates(payload.reject { |k, _v| k == 'request_id' })
+        if invalid.empty?
+          result = generate_ondemand_request(payload)
+        else
+          result[:bad_templates] = invalid
+          invalid.each do |bad_template|
+            metrics.increment('ondemandrequest.invalid.' + bad_template)
+          end
+          status 404
+        end
+      else
+        metrics.increment('ondemandrequest.invalid.unknown')
+        status 404
+      end
+
+      JSON.pretty_generate(result)
+    end
+
+    get "#{api_prefix}/ondemandvm/:requestid/?" do
+      content_type :json
+
+      status 404
+      result = check_ondemand_request(params[:requestid])
+
+      JSON.pretty_generate(result)
+    end
+
+    delete "#{api_prefix}/ondemandvm/:requestid/?" do
+      content_type :json
+      need_token! if Vmpooler::API.settings.config[:auth]
+
+      status 404
+      result = delete_ondemand_request(params[:requestid])
+
+      JSON.pretty_generate(result)
+    end
+
     post "#{api_prefix}/vm/?" do
       content_type :json
       result = { 'ok' => false }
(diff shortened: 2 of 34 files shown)


docker/docker-compose.yml
```diff
@@ -17,7 +20,8 @@ services:
       - VMPOOLER_DEBUG=true # for use of dummy auth
       - VMPOOLER_CONFIG_FILE=/etc/vmpooler/vmpooler.yaml
       - REDIS_SERVER=redislocal
-      - LOGFILE=/dev/stdout
+      - LOGFILE=/dev/null
```
highb, line 23: No logs?

lib/vmpooler/generic_connection_pool.rb
```diff
@@ -15,35 +15,17 @@ def initialize(options = {}, &block)
         @metric_prefix = 'connectionpool' if @metric_prefix.nil? || @metric_prefix == ''
       end
 
-      if Thread.respond_to?(:handle_interrupt)
-        # MRI
```
highb, line 19: Heh, I remember we discussed this code being fishy when debugging the Jruby issue. Isn't it essentially unreachable?

highb:
I follow what's going on so far. It's good that you added a redis connection pool. Tried it out and it appears to work:
```
❯ http post http://localhost:4567/api/v1/ondemandvm centos-7-x86_64=1 X-Auth-Token:$TOKEN

HTTP/1.1 201 Created
Content-Length: 72
Content-Type: application/json
X-Content-Type-Options: nosniff

{
    "ok": true,
    "request_id": "c45595d7-ef8e-4b69-bd6e-4cb31d7ab86f"
}

~
❯ http get http://localhost:4567/api/v1/ondemandvm/c45595d7-ef8e-4b69-bd6e-4cb31d7ab86f
c45595d7-ef8e-4b69-bd6e-4cb31d7ab86f
HTTP/1.1 202 Accepted
Content-Length: 155
Content-Type: application/json
X-Content-Type-Options: nosniff

{
    "centos-7-x86_64": {
        "pending": "1",
        "ready": "0"
    },
    "ok": true,
    "ready": false,
    "request_id": "c45595d7-ef8e-4b69-bd6e-4cb31d7ab86f"
}

~
❯ http get http://localhost:4567/api/v1/ondemandvm/c45595d7-ef8e-4b69-bd6e-4cb31d7ab86f

HTTP/1.1 200 OK
Content-Length: 164
Content-Type: application/json
X-Content-Type-Options: nosniff

{
    "centos-7-x86_64": {
        "hostname": [
            "eclectic-vogue"
        ]
    },
    "ok": true,
    "ready": true,
    "request_id": "c45595d7-ef8e-4b69-bd6e-4cb31d7ab86f"
}
```

docker/docker-compose.yml
mattkirby: For some reason the logs get duplicated in my docker compose environment if I set this logfile to be stdout. I found that when I set it to be /dev/null that I got the log messages, but they were no longer duplicated. Wonky stuff. I can remove this if needed.

lib/vmpooler/generic_connection_pool.rb
mattkirby: Yes. With all of our tested MRI and jruby versions they respond to `:handle_interrupt` so I think it is past its time. We used to run old jruby with vmpooler, which is probably why this is here.

docker/docker-compose.yml
highb: That's pretty strange. I guess we don't really need logs in the docker compose environment beyond what goes to stdout already so it's probably fine.

lib/vmpooler/generic_connection_pool.rb
highb: 💀 RIP dead code 💀

spec/integration/api/v1/vm_hostname_spec.rb
```diff
@@ -146,6 +148,19 @@ def app()
...
+#         create_vm('testhost', redis)
+#
+#         set_vm_data('testhost', "checkout", (Time.now - (69*60*60)), redis)
+#         puts redis.hget("vmpooler__vm__testhost", 'checkout')
+#         put "#{prefix}/vm/testhost", '{"lifetime":"100"}'
+#         expect_json(ok = false, http = 400)
+#
+#         vm = fetch_vm('testhost')
+#         expect(vm['lifetime']).to be_nil
+#       end
```
genebean, line 163: What's with this being commented out vs removed?

mattkirby: > It looks like a ton of this is variations on a theme... it felt like I was seeing the same change over and over. I assume this is just from where you were swapping out the redis stuff... is that accurate?

Yes. Implementing the redis connection pool required a ton of changes, so a large portion of the changes are changing how we are calling redis so we are utilizing the connection pool.

mattkirby: > What's with this being commented out vs removed?

That is a failed rebase on my part. I brought in a test that was removed. I had commented it out while I could not figure out why it was not working until I later realized that it was a rebase conflict that I fixed that brought it back, after a commit had removed the test along with a change. I have a change coming that will remove this.

genebean:
Looking good :)

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,85 @@ def sync_pool_templates
...
+        result['message'] = "requested amount of instances exceeds the maximum #{config['max_ondemand_instances_per_request']}"
+        status 403
+        return result
+      end
+
+      score = Time.now.to_i
+      request_id = payload['request_id']
+      request_id ||= generate_request_id
+      result['request_id'] = request_id
+
```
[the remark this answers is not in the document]
sbeaulie: Also I don't know if it would make sense eventually but my `POST` code does allow for a logic flow where the answer has `ready` true and the hostnames are returned. I see that code path does not exist in your code (yet) but would it be possible in the future. Say we do a POST, vmpooler sees the requested VMs are already available and it gives them out?

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,85 @@ def sync_pool_templates
...
       template_configs = backend.hgetall('vmpooler__config__template')
       template_configs&.each do |poolname, template|
-          if pool_index.include? poolname
-            unless pools[pool_index[poolname]]['template'] == template
-              pools[pool_index[poolname]]['template'] = template
-            end
-          end
+        next unless pool_index.include? poolname
+
+        pools[pool_index[poolname]]['template'] = template unless pools[pool_index[poolname]]['template'] == template
```
highb, line 312: What's the original logic behind this? Is doing an assignment to this hash expensive somehow?

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,85 @@ def sync_pool_templates
...
       poolsize_configs = backend.hgetall('vmpooler__config__poolsize')
       poolsize_configs&.each do |poolname, size|
-          if pool_index.include? poolname
-            unless pools[pool_index[poolname]]['size'] == size.to_i
-              pools[pool_index[poolname]]['size'] == size.to_i
-            end
-          end
+        next unless pool_index.include? poolname
+
+        pools[pool_index[poolname]]['size'] == size.to_i unless pools[pool_index[poolname]]['size'] == size.to_i
```
highb, line 322: Similar question here to above, this seems very obscured to me.

highb:
Testing it out on vmpooler-dev.k8s, I'm still seeing the on demand pools reported as empty from the status endpoint:
```
❯ http get http://vmpooler-dev.k8s.infracore.puppet.net/api/v1/status centos-7-x86_64=1 X-Auth-Token:$TOKEN
...
    "status": {
        "empty": [
            "redhat-7-x86_64",
            "debian-9-x86_64",
            "centos-8-x86_64",
            "ubuntu-2004-x86_64"
        ],
        "message": "Found 4 empty pools.",
        "ok": false,
        "uptime": 2859.3
    }
}
```

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,85 @@ def sync_pool_templates
...
+
+        return true
+      end
+      false
+    end
+
+    def generate_ondemand_request(payload)
+      result = { 'ok': false }
+
+      if too_many_requested?(payload.reject { |k, _v| k == 'request_id' })
```
highb, line 348: This `payload.reject` to format the payload is done twice in this method, maybe it would be easier to read the intent of what it's meant to be doing if this were a variable like `payload_request` ?

lib/vmpooler/api/v1.rb
mattkirby: The configured template in redis may be different than the template configured in the configuration file on disk for vmpooler because the value provided via the configuration API takes precedence. This was added to ensure that the API got the expected data in the configuration at startup time.

lib/vmpooler/api/v1.rb
mattkirby: sync_pool_sizes is similar to syncing pool templates. If the pool size differs in the configuration file and in redis (from a previous API request changing the pool size) then on startup vmpooler would display the value in the configuration file. This method updates the configuration to reflect that the value loaded into redis take precedence over the value configured in the configuration file.

lib/vmpooler/api/v1.rb
highb: But wouldn't just assigning the template accomplish the same thing? I guess I'm not understanding why the conditional checking that the values aren't the same is necessary, as if `pools[pool_index[poolname]]['template'] == template` then the assignment is a no-op (maybe this is my misunderstanding), and if `pools[pool_index[poolname]]['template'] != template` then assignment is fine.

lib/vmpooler/api/v1.rb
mattkirby: Yes, that makes sense. I have pushed an update to move this assignment to a variable called `requested_instances`. Let me know if you think that is unclear.

lib/vmpooler/api/v1.rb
highb: This one is especially confusing because it's doing equality not assignment.

lib/vmpooler/api/v1.rb
highb: I think that makes the intent clearer. Thanks!

lib/vmpooler/api/v1.rb
mattkirby: Yes, I think it would. I have adjusted it. Let me know if I misunderstood your suggestion.

lib/vmpooler/api/v1.rb
mattkirby: Yes. I don't think that is what the intention was. I adjusted this so it is actually assigning data.

mattkirby:
> Testing it out on vmpooler-dev.k8s, I'm still seeing the on demand pools reported as empty from the status endpoint:
> 
> ```
> ❯ http get http://vmpooler-dev.k8s.infracore.puppet.net/api/v1/status centos-7-x86_64=1 X-Auth-Token:$TOKEN
> ...
>     "status": {
>         "empty": [
>             "redhat-7-x86_64",
>             "debian-9-x86_64",
>             "centos-8-x86_64",
>             "ubuntu-2004-x86_64"
>         ],
>         "message": "Found 4 empty pools.",
>         "ok": false,
>         "uptime": 2859.3
>     }
> }
> ```

That is not something I had tested out. Some additions will be needed to address this.

lib/vmpooler/api/v1.rb
highb: Yep, the changes in https://github.com/puppetlabs/vmpooler/pull/375/commits/f596a989a5acbe187ed0c905e24080939a826994 make way more sense to me. :)

mattkirby:
> > Testing it out on vmpooler-dev.k8s, I'm still seeing the on demand pools reported as empty from the status endpoint:
> > ```
> > ❯ http get http://vmpooler-dev.k8s.infracore.puppet.net/api/v1/status centos-7-x86_64=1 X-Auth-Token:$TOKEN
> > ...
> >     "status": {
> >         "empty": [
> >             "redhat-7-x86_64",
> >             "debian-9-x86_64",
> >             "centos-8-x86_64",
> >             "ubuntu-2004-x86_64"
> >         ],
> >         "message": "Found 4 empty pools.",
> >         "ok": false,
> >         "uptime": 2859.3
> >     }
> > }
> > ```
> 
> That is not something I had tested out. Some additions will be needed to address this.

I was able to address this by ensuring that the existing logic in status checks to see if the configured pool size is 0. If so, it does not report the pool as empty, or has having an issue.

highb:
Yep, that appears to have fixed it!

lib/vmpooler/api/v1.rb
```diff
@@ -764,6 +932,79 @@ def invalid_pool(payload)
...
+        platform_parts = request_hash['requested'].split(',')
+        platform_parts.each do |platform|
+          pool_alias, pool, _count = platform.split(':')
+          instances = backend.smembers("vmpooler__#{request_id}__#{pool_alias}__#{pool}")
+          result[pool_alias] = { 'hostname': instances }
+        end
+        result['domain'] = config['domain'] if config['domain']
+        status 200
+      elsif request_hash['status'] == 'failed'
+        result['ready'] = false
```
sbeaulie, line 959: this is already false from ln 944
```suggestion
```

lib/vmpooler/api/v1.rb
```diff
@@ -689,6 +802,61 @@ def sync_clone_targets
...
+              metrics.increment('ondemandrequest.invalid.' + bad_template)
+            end
+            status 404
+          end
+        else
+          metrics.increment('ondemandrequest.invalid.unknown')
+          status 404
+        end
+      rescue JSON::ParserError
+        status 400
```
sbeaulie, line 831: Can you please add this to the docs/API.md possible status codes

lib/vmpooler/api/v1.rb
mattkirby: Good point. This has been removed.

lib/vmpooler/api/v1.rb
mattkirby: Yes, I missed this one. I have added it to the docs/API.md.

mattkirby:
> 👍 Seems good to go; do you want to squash your commits? :shipit:

Squashed!

lib/vmpooler/api/v1.rb
```diff
@@ -764,6 +932,78 @@ def invalid_pool(payload)
...
+      end
+
+      result['request_id'] = request_id
+      result['ready'] = false
+      result['ok'] = true
+      status 202
+
+      if request_hash['status'] == 'ready'
+        result['ready'] = true
+        platform_parts = request_hash['requested'].split(',')
```
sbeaulie, line 950:
```suggestion
        platform_parts = request_hash['requested'].split(',')
        platform_parts.each do |platform|
          pool_alias, pool, _count = platform.split(':')
          instances = backend.smembers("vmpooler__#{request_id}__#{pool_alias}__#{pool}")
```

I saw variations of this in 4 places, where we deconstruct I wonder if a method that accepts a block could DRY it? And we could add some input data validation around the `split()` I'm thinking what if `split(":")` has nil for `pool_alias`, `pool`
Is it possible for `backend.smembers` to also return nil?

sbeaulie: 2. https://github.com/puppetlabs/vmpooler/pull/375/files#diff-51d4606238cfb066704749b609c2d728R965
3. https://github.com/puppetlabs/vmpooler/pull/375/files#diff-51d4606238cfb066704749b609c2d728R993
4. https://github.com/puppetlabs/vmpooler/pull/375/files#diff-9ca942eee48c471766a71aa4fd1df851R1477
5. https://github.com/puppetlabs/vmpooler/pull/375/files#diff-9ca942eee48c471766a71aa4fd1df851R1523

lib/vmpooler/api/v1.rb
```diff
@@ -395,7 +508,7 @@ def sync_clone_targets
           end
 
           # for backwards compatibility, include separate "empty" stats in "status" block
-          if ready == 0
+          if ready == 0 && max != 0
```
sbeaulie, line 511: Is that the only place where we had to check that a pool is `ondemand` based on the max value being 0?
I'd like to encapsulate that in a utility method that makes it clear the relationship between being ondemand and max ==0 in case in the future we change what criteria that is. If this is the only place then its not so bad and I can be convinced to keep it that way.

I was thinking
```suggestion
          if ready == 0 && !is_ondemand?
```

with that method checking if the pool has its max == 0. Or whatever logical inverse is if that makes sense

mattkirby: I believe it shows up once in API, and once in pool manager, because we perform a similar check to avoid those messages in each location.

lib/vmpooler/api/v1.rb
mattkirby: Maybe a method that handles grabbing the string, splitting it up and returning the vars to be assigned?
`backend.smembers` returns an empty array when the set is empty.

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,87 @@ def sync_pool_templates
...
+      false
+    end
+
+    def generate_ondemand_request(payload)
+      result = { 'ok': false }
+
+      requested_instances = payload.reject { |k, _v| k == 'request_id' }
+      if too_many_requested?(requested_instances)
+        result['message'] = "requested amount of instances exceeds the maximum #{config['max_ondemand_instances_per_request']}"
+        status 403
```
jcoconnor, line 351: Should add ```ondemandrequest.toomanyrequests.``` here

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,87 @@ def sync_pool_templates
...
+      end
+
+      score = Time.now.to_i
+      request_id = payload['request_id']
+      request_id ||= generate_request_id
+      result['request_id'] = request_id
+
+      if backend.exists("vmpooler__odrequest__#{request_id}")
+        result['message'] = "request_id '#{request_id}' has already been created"
+        status 409
```
jcoconnor, line 362: Maybe should add ```ondemandrequest.duplicaterequests.``` here

lib/vmpooler/api/v1.rb
```diff
@@ -245,36 +307,87 @@ def sync_pool_templates
...
+
+      return result unless backend.zadd('vmpooler__provisioning__request', score, request_id)
+
+      backend.hset("vmpooler__odrequest__#{request_id}", 'requested', platforms_string)
+      if Vmpooler::API.settings.config[:auth] and has_token?
+        backend.hset("vmpooler__odrequest__#{request_id}", 'token:token', request.env['HTTP_X_AUTH_TOKEN'])
+        backend.hset("vmpooler__odrequest__#{request_id}", 'token:user',
+                     backend.hget('vmpooler__token__' + request.env['HTTP_X_AUTH_TOKEN'], 'user'))
+      end
+
```
jcoconnor, line 383: Should add ```ondemandrequest.success.``` here  (or request success?)

lib/vmpooler/api/v1.rb
mattkirby: I looked at this a little more and am not quite sure the best way to address this. We could have a method return the pool_alias, pool and count for each of the requested platforms in the requested string, and then take the action on that returned data. I'm not sure if it clarifies things much.
I didn't see a  clear  way  to make this better with a block. Did you have a specific idea  in mind for how to improve this? I'm open to ideas, but did  not  see  a clear path to improvement.

lib/vmpooler/api/v1.rb
mattkirby: Yes, I think that sounds good, or perhaps `toomanyrequested`

lib/vmpooler/api/v1.rb
mattkirby: Yes, that makes sense.

lib/vmpooler/api/v1.rb
mattkirby: Yes, I think `ondemandrequest.success` makes sense.

lib/vmpooler/api/v1.rb
sbeaulie: yes a utility method like
```
    def split_platform_pool_count(requested, &block)
      requested_platforms = requested.split(',')
        requested_platforms.each do |platform|
          platform_alias, pool, count = platform.split(':')
         #TODO data validation on platform_alias, pool, count if nil
          yield platform_alias, pool, count
        end
    end
```
could then be called in the first example like:
```
split_platform_pool_count(request_hash['requested']) do |platform_alias,pool,count|
            instances = backend.smembers("vmpooler__#{request_id}__#{pool_alias}__#{pool}")
          result[pool_alias] = { 'hostname': instances }
end
```
the second example in manager becomes
```
split_platform_pool_count(request_hash['requested']) do |platform_alias,pool,count|
  platform_alias, pool, count = platform.split(':')
  pools_filled = redis.scard("vmpooler__#{request_id}__#{platform_alias}__#{pool}")
  throw :request_not_ready unless pools_filled.to_i == count.to_i
end
```
etc

obviously naming can be improved and do we have a natural place for that utility method?

mattkirby: We do not. API has its own helpers class. So without introducing something new, like a generic helpers class, I would have to duplicate the utility method in API and manager.

PR #377, opened 2020-05-29
(POOLER-166) Check for stale dns records

sbeaulie:
Before this change we checked that the new vm name would not conflict with an existing vm already running and under the vmpooler management in redis.
We now also check if a DNS query returns an existing IP, which could happen if there are stale DNS record (or some other process registered the vm name already, like another vmpooler?). In that case we skip that name and try again.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -301,14 +301,14 @@ module Vmpooler
       hostname_retries = 0
       max_hostname_retries = 3
       while hostname_retries < max_hostname_retries
-        hostname, available = generate_and_check_hostname(pool_name)
+        hostname, hostname_available = generate_and_check_hostname(pool_name)
         domain = $config[:config]['domain']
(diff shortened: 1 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -295,22 +296,46 @@ def generate_and_check_hostname(_pool_name)
...
+      # generate hostname that is not already in use in vmpooler
+      # also check that no dns record already exists
       hostname_retries = 0
       max_hostname_retries = 3
       while hostname_retries < max_hostname_retries
         hostname, available = generate_and_check_hostname(pool_name)
-        break if available
+        domain = $config[:config]['domain']
+        dns_ip, dns_available = check_dns_available(hostname, domain)
+        break if available && dns_available
```
rooneyshuman, line 307: with the addition of the dns availability check, it might be good to rename `available` to `hostname_available` to differentiate what each of these variables check for

PR #379, opened 2020-06-05
Rescue and warn when graphite connection cannot be opened

mattkirby:
This commit updates vmpooler graphite stats behavior when a connection cannot be opened to the graphite server to warn instead of allowing the full backtrace to be output to stderr. Without this change a backtrace is output to stderr when a graphite connection fails.

### DIFF
--- a/lib/vmpooler/graphite.rb
+++ b/lib/vmpooler/graphite.rb
@@ -35,6 +35,8 @@ module Vmpooler
           socket.close
         end
       end
+    rescue Errno::EADDRNOTAVAIL => e
+      warn "Could not assign address to graphite server #{server}: #{e}"
     rescue StandardError => e
       warn "Failure logging #{path} to graphite server [#{server}:#{port}]: #{e}"
     end


lib/vmpooler/graphite.rb
```diff
@@ -35,6 +35,8 @@ def log(path, value)
           socket.close
         end
       end
+    rescue Errno::EADDRNOTAVAIL => e
```
highb, line 38: Are there any other `Errno` we'd want to catch here? `Errno::EACCES` and `Errno::EADDRINUSE` seem like those are probably fatal errors but would `Errno::EFAULT` ever be non-fatal?

mattkirby: Quite possibly. EADDRNOTAVAIL caught my attention because it's what our instance is yelling about now when it cannot get a connection to graphite. I'm not sure when we would encounter the others, but am not opposed to adding if you'd like to see those too.

highb: Problem is, I'm not entirely sure which ones would be useful to add, and which would hide a fatal problem. 😆  Maybe it's better to just ship this for now and figure out additional errors to catch later...

PR #383, opened 2020-06-11
(POOLER-174) Reduce duplicate of on demand code introduced in POOLER-158

sbeaulie:
added leftover suggestions from #375
refactored every parsing of request of type 'pool_alias:pool:count' into a
utility class, that is used by pool_manager and the api v1 class
added metrics to the ondemand request generation


mattkirby:
Thanks for working on this one and addressing these remaining items @sbeaulie !

sbeaulie:
Rubocop failures, fixing linting

sbeaulie:
thanks

PR #385, opened 2020-07-03
(POOLER-181) Separate Dashboard from api

jcoconnor:
Allow the dashboard to run separately from the API, so that the
dashboard can be scaled appropriately.

Note - this implementation actually co-runs the api in the dashboard
instance to re-use the api code for the dashboard endpoints and other
v1 endpoints collecting the operating information.

### DIFF
--- a/bin/vmpooler
+++ b/bin/vmpooler
@@ -16,20 +16,21 @@ metrics = Vmpooler::Metrics.init(logger, config)
 
 torun_threads = []
 if ARGV.count == 0
-  torun = %i[api manager]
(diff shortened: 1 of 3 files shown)


genebean:
@jcoconnor IMO, the dashboard should not run an instance of the API. It should reach out to wherever the API is already deployed as it should simply consume the API like everything else does.

jcoconnor:
Closing - won't happen before I leave 😭  and @genebean comment is correct - there is a better way to do it.

PR #389, opened 2020-07-23
Fix vmpooler folder purging

mattkirby:
This commit updates folder purging references to ensure that provider
name references are referring to the named provider, rather than the
provider type. Without this change folder purging fails because it
cannot identify target folders.


mattkirby:
I fixed up the tests here, this should be ready to go.

PR #390, opened 2020-07-29
(MAINT) Clarity refactor of Prom Stats code

jcoconnor:
Introducing the Prometheus Stats code into ABS showed that the Clarity
could be improved a bit with better variable naming, some refactoring
to reduce repitition and documenting the Metrics table itself.

Filtering these changes back to the vmpooler code base.

### DIFF
--- a/lib/vmpooler/metrics/promstats.rb
+++ b/lib/vmpooler/metrics/promstats.rb
@@ -5,7 +5,7 @@ require 'prometheus/client'
 module Vmpooler
   class Metrics
     class Promstats < Metrics
-      attr_reader :prefix, :endpoint, :metrics_prefix
+      attr_reader :prefix, :prometheus_endpoint, :prometheus_prefix
 
       # Constants for Metric Types
       M_COUNTER   = 1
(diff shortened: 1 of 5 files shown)


lib/vmpooler/metrics/promstats.rb
```diff
@@ -159,82 +146,72 @@ def vmpooler_metrics_table
...
             docstring: 'vmpooler machine migrated to',
-            prom_metric_prefix: "#{@metrics_prefix}_migrate_to",
             param_labels: %i[host_name]
           },
           api_vm: {
             mtype: M_COUNTER,
             torun: %i[api],
             docstring: 'Total number of HTTP request/sub-operations handled by the Rack application under the /vm endpoint',
-            prom_metric_prefix: "#{@metrics_prefix}_http_requests_vm_total",
+            prom_metric_id: 'http_requests_vm_total',
```
genebean, line 161: Why is this one different from all the others?

lib/vmpooler/metrics/promstats.rb
```diff
@@ -315,9 +295,9 @@ def find_metric(label)
           metric_subkey = sublabels.shift.to_sym
           raise("Invalid Metric #{metric_key}_#{metric_subkey} for #{label}") unless metric[:metric_suffixes].key? metric_subkey.to_sym
 
-          metric[:metric_name] = "#{metric[:prom_metric_prefix]}_#{metric_subkey}"
+          metric[:metric_name] = "#{@prometheus_prefix}_#{metric[:prom_metric_id]}_#{metric_subkey}"
         else
-          metric[:metric_name] = metric[:prom_metric_prefix]
+          metric[:metric_name] = "#{@prometheus_prefix}_#{metric[:prom_metric_id]}"
```
genebean, line 300: It looks like you intend to replace `prom_metric_prefix` with `prom_metric_id` but all but one entry in the table above simply removes `prom_metric_prefix` instead.

lib/vmpooler/metrics/promstats.rb
jcoconnor: Thats actually a good question - should I should change it in the ```.increment``` statements``` ?
Probably 😀

jcoconnor: Fixed the one anomaly and simplified this.

lib/vmpooler/metrics/promstats.rb
jcoconnor: Have simplified it all a bit.

lib/vmpooler/metrics/promstats.rb
```diff
@@ -23,9 +23,9 @@ class Promstats < Metrics
       @p_metrics = {}
 
       def initialize(logger, params = {})
-        @prefix = params['prefix'] || 'vmpooler'
-        @metrics_prefix = params['metrics_prefix'] || 'vmpooler'
-        @endpoint = params['endpoint'] || '/prometheus'
+        @prefix = params[:prefix] || 'vmpooler'
+        @prometheus_prefix = params[:prometheus_prefix] || 'vmpooler'
+        @prometheus_endpoint = params[:prometheus_endpoint] || '/prometheus'
```
genebean, line 28: I question changing line 26 to a symbol - that will break existing configuration and does not seem to correspond to how other metrics providers are configured.

jcoconnor: Yeah - have fixed the test params spec issue so reverted back to the standard way.

PR #400, opened 2020-09-12
Add additional data to spans in api/v1.rb

genebean:
I have tested this locally and it all seems to work.

This PR is built atop #471


lib/vmpooler/api/v1.rb
```diff
@@ -109,398 +113,462 @@ def evaluate_template_aliases(template, count)
...
+            metrics.increment("usage_job_component.#{job_name}.#{component_to_test}.#{operation}.#{poolname}")
+          end
+        rescue StandardError => e
+          puts 'd', "[!] [#{poolname}] failed while evaluating usage labels on '#{vmname}' with an error: #{e}"
+          span.add_event('error', attributes: {
+            'error.type' => 'Vmpooler::API::V1.update_user_metrics',
+            'error.message' => "[#{poolname}] failed while evaluating usage labels on '#{vmname}' with an error: #{e}"
+          })
+        end
       end
```
genebean, line 326: @sbeaulie Rubocop thows the warning below now that I have added the tracer bit and a `begin` to this method... is there a better way to do this or should I add a control comment to tell Rubocop this is okay?

```ruby
lib/vmpooler/api/v1.rb:301:15: W: Lint/NonLocalExitFromIterator: Non-local exit from iterator, without return value. next, break, Array#find, Array#any?, etc. is preferred.
              return
              ^^^^^^
```

sbeaulie (approved):

sbeaulie (approved):

PR #405, opened 2020-09-18
Adding make to the other two Dockerfiles

genebean:
In a previous PR I added make to Dockerfiles_local but didn't realize it also needed to be in the other two Dockerfiles... this fixes that


sbeaulie (approved):

sbeaulie:
maybe useful to comment what make is used for, is it only for `thrift` ?

PR #422, opened 2020-10-23
(maint) Speedup the tagging method

sbeaulie:
While looking at the instrumentation data for the ABS queue processor,
I noticed a lot of time spent in the HTTP PUT method, which in the code
was easy to isolate, as it is only used via the vmpooler tagging functions
ie the API /vm/foobar/ with 'tag' key-value pairs.
While I'm not sure the original hset() make sense to me, there was an easy
way to speed them up by using pipelined. I would expect a very good speed
increase with this turned on.

### DIFF
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -114,11 +114,13 @@ module Vmpooler
       end
 
       def export_tags(backend, hostname, tags)
-        tags.each_pair do |tag, value|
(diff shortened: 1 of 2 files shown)


genebean (approved):
LGTM but I don't know enough about this to fully understand the implications

mattkirby (approved):

mattkirby:
This was tested on vmpooler-dev and all works well.

PR #423, opened 2020-10-24
(POOLER-192) Use Rubocop 1.0

rooneyshuman:
Fixes errors resulting from using Rubocop's new stable version v1.0 and updates the gem.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -108,7 +108,7 @@ module Vmpooler
       $logger.log('d', "[!] [#{pool}] '#{vm}' no longer exists. Removing from pending.")
     end
 
-    def fail_pending_vm(vm, pool, timeout, redis, exists = true)
+    def fail_pending_vm(vm, pool, timeout, redis, exists: true)
       clone_stamp = redis.hget("vmpooler__vm__#{vm}", 'clone')
 
       time_since_clone = (Time.now - Time.parse(clone_stamp)) / 60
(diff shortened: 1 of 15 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -867,16 +868,12 @@ def sleep_with_wakeup_events(loop_delay, wakeup_period = 5, options = {})
 
             if options[:clone_target_change]
               clone_target = redis.hget('vmpooler__config__clone_target}', options[:poolname])
-              if clone_target
-                break unless clone_target == initial_clone_target
-              end
+              break if clone_target && !clone_target == initial_clone_target
```
sbeaulie: unless clone_target is a Boolean, this is not the same. ! on a string returns false.
```suggestion
              break if clone_target && clone_target != initial_clone_target
```

sbeaulie (changes requested):

lib/vmpooler/providers/vsphere.rb
```diff
@@ -418,15 +416,15 @@ def set_network_device(datacenter_name, template_vm_network_device, network_name
           # Determine network device type
           # All possible device type options here: https://vdc-download.vmware.com/vmwb-repository/dcr-public/98d63b35-d822-47fe-a87a-ddefd469df06/2e3c7b58-f2bd-486e-8bb1-a75eb0640bee/doc/vim.vm.device.VirtualEthernetCard.html
           network_device =
-            if template_vm_network_device.is_a? RbVmomi::VIM::VirtualVmxnet2
```
sbeaulie, line 421: we need to check if there are subclasses that we also want to accept here? for example if `RbVmomi::VIM::VirtualE1000` has a subclass, `is_a?(RbVmomi::VIM::VirtualE1000)` would return `true` for the subclass too

lib/vmpooler/providers/vsphere.rb
```diff
@@ -49,9 +51,7 @@ def name
         end
 
         def folder_configured?(folder_title, base_folder, configured_folders, whitelist)
-          if whitelist
-            return true if whitelist.include?(folder_title)
-          end
+          return true if whitelist&.include?(folder_title)
```
sbeaulie, line 54: what kind of magic is that!

rooneyshuman: the safe navigation operator is probably the coolest thing I saw from this rubocop change

sbeaulie (approved):

lib/vmpooler/providers/vsphere.rb
rooneyshuman: Those network device classes don't have any subclasses. They are the exact class for the virtual network device objects

lib/vmpooler/providers/vsphere.rb
jcoconnor: That does look rather cool.

jcoconnor (approved):

PR #434, opened 2020-11-06
(POOLER-193) Mark checked out VM as active

mattkirby:
This change sets a VM as running in redis as soon as it is checked out. Without this change when allocated several instances it is possible for a machine that has been allocated for a checkout, but not yet marked as active, to be identified as running when it should not be, which was added in POOLER-191. Without this change a machine may be destroyed during checkout by pool_manager if there are several instances being allocated.

Additionally, redis multi is added for vm checkout operations to minimize the round trips to redis during a checkout operation. Without this addition each VM checkout causes several redis interactions.

### DIFF
--- a/lib/vmpooler/api/helpers.rb
+++ b/lib/vmpooler/api/helpers.rb
@@ -18,7 +18,7 @@ module Vmpooler
 
       def validate_token(backend)
         if valid_token?(backend)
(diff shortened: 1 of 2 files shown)


jcoconnor:
@mattkirby Think you are falling foul on the same pre-deploy issue I did.

genebean (approved):

jcoconnor (approved):

PR #453, opened 2021-06-04
Pooler 194

tanisha-payne:
Not sure how to pass a dummy auth token into restart_spec.rb


yachub:
Will revisit this ticket at a later date.

PR #455, opened 2021-08-11
(POOLER-176) Add Operation Label to User Metric

yachub:
This adds an "operation" label to the user metrics and moves incrementing from the manager to api, so that the user metrics show when resources are allocated, as well as destroyed. Previously, user metrics were only updated upon destroying a resource.

I think its better suited to increment the metric as part of the api instead of the pool_manger, because it's expected to do so when a user successfully checks out or deletes a VM, but can be problematic when doing so in the provider since it can clone VMs before actually being checked out by a user.


lib/vmpooler/api/v1.rb
```diff
@@ -235,6 +236,47 @@ def atomically_allocate_vms(payload)
...
+        build_metadata_parts = url_parts[3]
+        component_to_test = component_to_test('RMM_COMPONENT_TO_TEST_NAME', build_metadata_parts)
+
+        metrics.increment("usage_jenkins_instance.#{jenkins_instance}.#{value_stream}.#{operation}.#{poolname}")
+        metrics.increment("usage_branch_project.#{branch}.#{project}.#{operation}.#{poolname}")
+        metrics.increment("usage_job_component.#{job_name}.#{component_to_test}.#{operation}.#{poolname}")
+      end
+    rescue StandardError => e
+      puts 'd', "[!] [#{poolname}] failed while evaluating usage labels on '#{vmname}' with an error: #{e}"
+      raise
```
sbeaulie, line 277: you could consider not re-raising it here since if it fails for some reason like an unexpected `nil` value it would afect only metrics but the operation to atomically_allocate_vms would not be affected.

lib/vmpooler/metrics/promstats.rb
```diff
@@ -165,33 +165,33 @@ def vmpooler_metrics_table
...
+            docstring: 'Number of pool instances and the operation performed by Jenkins instances',
+            param_labels: %i[jenkins_instance value_stream operation poolname]
           },
           usage_branch_project: {
             mtype: M_COUNTER,
-            torun: %i[manager],
-            docstring: 'Pools by branch/project usage',
-            param_labels: %i[branch project poolname]
+            torun: %i[api],
+            docstring: 'Number of pool instances and the operation performed by Litmus jobs by Jenkins branch/project',
```
sbeaulie, line 187:
```suggestion
            docstring: 'Number of pool instances and the operation performed by Jenkins branch/project',
```

lib/vmpooler/metrics/promstats.rb
```diff
@@ -165,33 +165,33 @@ def vmpooler_metrics_table
...
+            param_labels: %i[branch project operation poolname]
           },
           usage_job_component: {
             mtype: M_COUNTER,
-            torun: %i[manager],
-            docstring: 'Pools by job/component usage',
-            param_labels: %i[job_name component_to_test poolname]
+            torun: %i[api],
+            docstring: 'Number of pool instances and the operation performed by Litmus jobs Jenkins by job/component',
```
sbeaulie, line 193:
```suggestion
            docstring: 'Number of pool instances and the operation performed by Jenkins by job/component',
```

lib/vmpooler/api/v1.rb
yachub: Good call, no reason to raise an error and cause a VM not to checkout because of a failure to update metrics. Thanks

sbeaulie (approved):

PR #456, opened 2021-08-12
[WIP] Allow loading pools from a directory of files

genebean:
This allows for specifying a directory containing yaml files with one or more array entries inside where each array entry is a pool definition. The reason for this addition is to facilitate having one ConfigMap per pool in Kubernetes.


lib/vmpooler.rb
```diff
@@ -46,6 +46,15 @@ def self.config(filepath = 'vmpooler.yaml')
...
         end
       end
+      parsed_config[:config]['pools_from_dir'] = ENV['POOLS_FROM_DIR'] if ENV['POOLS_FROM_DIR']
+      if parsed_config[:config]['pools_from_dir'] && Dir.exist?(parsed_config[:config]['pools_from_dir'])
+        pools_from_dir = []
+        Dir["#{parsed_config[:config]['pools_from_dir']}/*.yaml"].sort.each do |file_name|
+          parsed_pool = Yaml.load_file(file_name)
+          pools_from_dir += parsed_pool
+        end
+        parsed_config[:pools] = pools_from_dir unless pools.empty?
```
sbeaulie, line 56: what's pools.empty?
```suggestion
        parsed_config[:pools] = pools_from_dir unless pools_from_dir.empty?
```

lib/vmpooler.rb
```diff
@@ -46,6 +46,15 @@ def self.config(filepath = 'vmpooler.yaml')
           parsed_config.merge!(extra_config)
         end
       end
+      parsed_config[:config]['pools_from_dir'] = ENV['POOLS_FROM_DIR'] if ENV['POOLS_FROM_DIR']
+      if parsed_config[:config]['pools_from_dir'] && Dir.exist?(parsed_config[:config]['pools_from_dir'])
+        pools_from_dir = []
+        Dir["#{parsed_config[:config]['pools_from_dir']}/*.yaml"].sort.each do |file_name|
+          parsed_pool = Yaml.load_file(file_name)
+          pools_from_dir += parsed_pool
```
sbeaulie, line 54: turns out load_file returns `false` if the file exists but fails to parse as yaml. since false in your array parsed_pool would not work i suggest
```suggestion
          pools_from_dir += parsed_pool if parsed_pool
```

genebean: Good call. Adding

lib/vmpooler.rb
genebean: Thanks!

PR #458, opened 2021-08-25
(DIO-541) Fix jenkins and user usage metrics

yachub:


genebean (commented):
Code change seems fine but I request that you look into adding the unit test to the api instead of just killing it off totally. If that is practical, at least test coverage would not be reduced.

genebean (approved):

PR #459, opened 2021-09-14
(DIO-2621) Make LDAP encryption configurable

genebean:
Prior to this, the encryption settings for LDAP auth were hard coded to start_tls on port 389 with TLSv1. These are still the defaults, as insecure as they are, so as to not break existing users. This change facilitates replacing the defaults so that simple_tls over port 636 via TLS1.2 can be used.


mattkirby (approved):

genebean:
Deployed to test and validated that using the config below allows for successful token retrieval:

```yaml
:auth:
  provider: 'ldap'
  :ldap:
    host: 'ldap.example.com'
    port: 636
    encryption:
      :method: :simple_tls
      :tls_options:
```

PR #467, opened 2021-12-08
Move vsphere specific methods out of vmpooler

sbeaulie:
vmpooler has the vsphere provider taken out, moving some vsphere related
methods to the provider:

1) pool_folders
2) get_base_folders

At the same time renaming some configuration and code items
to remove harmful terminology.
purge_unconfigured_folders DEPRECATED, use purge_unconfigured_resources
folder_whitelist DEPRECATED, use resources_allowlist
the above configuration items are still supported but will be removed in
the next major version.
base class method purge_unconfigured_folders was renamed to purge_unconfigured_resources
and requires the equivalent change in the provider classes

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -169,6 +169,7 @@ This can also be set per pool.
 (optional; default: true)
 
 ### PURGE\_UNCONFIGURED\_FOLDERS
+
 Deprecated, see PURGE\_UNCONFIGURED\_RESOURCES
 
 ### PURGE\_UNCONFIGURED\_RESOURCES
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -169,10 +169,12 @@ This can also be set per pool.
 (optional; default: true)
 
 ### PURGE\_UNCONFIGURED\_FOLDERS
+Deprecated, see PURGE\_UNCONFIGURED\_RESOURCES
 
-Enable purging of VMs and folders detected within the base folder path that are not configured for the provider
-Only a single layer of folders and their child VMs are evaluated from detected base folder paths
-A base folder path for 'vmpooler/redhat-7' would be 'vmpooler'
+### PURGE\_UNCONFIGURED\_RESOURCES
+
+Enable purging of VMs (and other resources) detected within the provider that are not explicitly configured.
+Implementation is provider-dependent
 When enabled in the global configuration then purging is enabled for all providers
 Expects a boolean value
 (optional; default: false)
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -81,16 +81,13 @@ module Vmpooler
     parsed_config[:config]['retry_factor']                       = string_to_int(ENV['RETRY_FACTOR']) if ENV['RETRY_FACTOR']
     parsed_config[:config]['create_folders']                     = true?(ENV['CREATE_FOLDERS']) if ENV['CREATE_FOLDERS']
     parsed_config[:config]['experimental_features']              = ENV['EXPERIMENTAL_FEATURES'] if ENV['EXPERIMENTAL_FEATURES']
-    parsed_config[:config]['purge_unconfigured_resources']       = ENV['PURGE_UNCONFIGURED_RESOURCES'] if ENV['PURGE_UNCONFIGURED_RESOURCES']
-    # ENV PURGE_UNCONFIGURED_FOLDERS deprecated, will be removed in version 3
-    if ENV['PURGE_UNCONFIGURED_FOLDERS']
-      puts '[!] [deprecation] rename ENV var \'PURGE_UNCONFIGURED_FOLDERS\' to \'PURGE_UNCONFIGURED_RESOURCES\''
-      # backwards compatibility
-      parsed_config[:config]['purge_unconfigured_resources']     = ENV['PURGE_UNCONFIGURED_FOLDERS']
-    end
     parsed_config[:config]['usage_stats']                        = ENV['USAGE_STATS'] if ENV['USAGE_STATS']
     parsed_config[:config]['request_logger']                     = ENV['REQUEST_LOGGER'] if ENV['REQUEST_LOGGER']
     parsed_config[:config]['create_template_delta_disks']        = ENV['CREATE_TEMPLATE_DELTA_DISKS'] if ENV['CREATE_TEMPLATE_DELTA_DISKS']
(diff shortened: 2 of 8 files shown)


docs/configuration.md
```diff
@@ -169,10 +169,12 @@ This can also be set per pool.
 (optional; default: true)
 
 ### PURGE\_UNCONFIGURED\_FOLDERS
+Deprecated, see PURGE\_UNCONFIGURED\_RESOURCES
```
genebean, line 172:
```suggestion
### PURGE\_UNCONFIGURED\_FOLDERS

Deprecated, see PURGE\_UNCONFIGURED\_RESOURCES
```

lib/vmpooler.rb
```diff
@@ -81,7 +81,9 @@ def self.config(filepath = 'vmpooler.yaml')
     parsed_config[:config]['retry_factor']                       = string_to_int(ENV['RETRY_FACTOR']) if ENV['RETRY_FACTOR']
     parsed_config[:config]['create_folders']                     = true?(ENV['CREATE_FOLDERS']) if ENV['CREATE_FOLDERS']
     parsed_config[:config]['experimental_features']              = ENV['EXPERIMENTAL_FEATURES'] if ENV['EXPERIMENTAL_FEATURES']
-    parsed_config[:config]['purge_unconfigured_folders']         = ENV['PURGE_UNCONFIGURED_FOLDERS'] if ENV['PURGE_UNCONFIGURED_FOLDERS']
+    parsed_config[:config]['purge_unconfigured_resources']       = ENV['PURGE_UNCONFIGURED_RESOURCES'] if ENV['PURGE_UNCONFIGURED_RESOURCES']
+    # ENV PURGE_UNCONFIGURED_FOLDERS deprecated, will be removed in version 3
+    parsed_config[:config]['purge_unconfigured_resources']       = ENV['PURGE_UNCONFIGURED_FOLDERS'] if ENV['PURGE_UNCONFIGURED_FOLDERS']
```
genebean, line 86: Would it be practical to log a message here with the depreciation message?

lib/vmpooler/pool_manager.rb
```diff
@@ -496,33 +496,16 @@ def purge_unused_vms_and_folders
...
-
-    def get_base_folders(folders)
-      base = []
-      folders.each do |_key, value|
-        base << value
-      end
-      base.uniq
-    end
-
     def purge_vms_and_folders(provider_name)
```
genebean, line 499: Does the name need to have `folders` removed from it?

lib/vmpooler/providers/base.rb
```diff
@@ -237,8 +237,15 @@ def get_target_datacenter_from_config(_provider_name)
           raise("#{self.class.name} does not implement get_target_datacenter_from_config")
         end
 
-        def purge_unconfigured_folders(_base_folders, _configured_folders, _whitelist)
-          raise("#{self.class.name} does not implement purge_unconfigured_folders")
+        def purge_unconfigured_resources(_whitelist)
```
genebean, line 240:
```suggestion
        def purge_unconfigured_resources(_allowlist)
```

lib/vmpooler/providers/base.rb
```diff
@@ -237,8 +237,15 @@ def get_target_datacenter_from_config(_provider_name)
...
+          raise("#{self.class.name} does not implement purge_unconfigured_resources")
+        end
+
+        # DEPRECATED if a provider does not implement the new method, it will hit this base class method
+        # and return a deprecation message
+        def purge_unconfigured_folders(_deprecated, _deprecated2, whitelist)
+          logger.log('s', '[!] purge_unconfigured_folders was renamed to purge_unconfigured_resources, please update your provider implementation')
+          purge_unconfigured_resources(whitelist)
```
genebean, line 248:
```suggestion
        def purge_unconfigured_folders(_deprecated, _deprecated2, allowlist)
          logger.log('s', '[!] purge_unconfigured_folders was renamed to purge_unconfigured_resources, please update your provider implementation')
          purge_unconfigured_resources(allowlist)
```

vmpooler.yaml.example
```diff
@@ -11,20 +11,21 @@
...
-#     To exclude 'vmpooler/myfolder' from being destroyed when the base path is 'vmpooler' you would specify 'myfolder' in the whitelist
-#     This option is only evaluated when 'purge_unconfigured_folders' is enabled
-#     Expects an array of strings specifying the whitelisted folders by name
+#   - folder_whitelist DEPRECATED, use resources_allowlist
+#   - resources_allowlist
+#     Specify items names that should be ignored when purging. See the provider's
+#     implementation for more details.
+#     This option is only evaluated when 'purge_unconfigured_resources' is enabled
+#     Expects an array of strings specifying the allowed items by name
 #     (optional; default: nil)
```
genebean, line 29: Since this will now vary by provider, should there even be a global one?

genebean (changes requested):

vmpooler.yaml.example
sbeaulie: there is no global one for resources_allowlist. Only the flag purge_unconfigured_resources is global (and overridable at the provider level)

genebean: I must have misread this section - I thought it was at the global level.

lib/vmpooler/pool_manager.rb
sbeaulie: sure, thats all internal and part of the main loop, so a remane should not affect anything, but there is a cascade of methods that need renamed, I'll do that now

genebean: Thanks. I think it will help with making the code easier for others to understand.

lib/vmpooler.rb
sbeaulie: I thought about that but there is no logger in that init class. In any case the code is using 'purge_unconfigured_resources' but still accepting the old ENV var. We could do a simple "puts" if that's sufficient?

lib/vmpooler/providers/base.rb
genebean: It looks like the whtielist stuff came back in somehow

lib/vmpooler/providers/base.rb
genebean: It looks like the whtielist stuff came back in somehow

lib/vmpooler.rb
genebean: I think that would work

genebean:
It looks like you might need a rebase on current main @sbeaulie

genebean (approved):

genebean (approved):

PR #468, opened 2021-12-09
Update testing.yml

sbeaulie:
since we use main as default branch now


genebean (approved):
Sorry I missed this when changing the default branch!

PR #470, opened 2021-12-10
Ensure all configured providers are loaded

genebean:
Prior to this, providers that should have been loaded per the provider_class key of the providers hash were not actually loaded.

### DIFF
--- a/lib/vmpooler/pool_manager.rb
+++ b/lib/vmpooler/pool_manager.rb
@@ -619,8 +619,21 @@ module Vmpooler
     # @return [Array] - a list of used providers from the config file, defaults to the default providers
     # ie. ["dummy"]
     def used_providers
-      pools = config[:pools] || []
-      @used_providers ||= (pools.map { |pool| pool[:provider] || pool['provider'] }.compact + default_providers).uniq
+      # create an array of provider classes based on the config
+      if config[:providers]
+        config_provider_names = config[:providers].keys
+        config_providers = config_provider_names.map do |config_provider_name|
+          if config[:providers][config_provider_name] && config[:providers][config_provider_name]['provider_class']
+            config[:providers][config_provider_name]['provider_class'].to_s
+          else
+            config_provider_name.to_s
+          end
+        end.compact.uniq
+      else
+        config_providers = []
+      end
+      # return the unique array of providers from the config and VMPooler defaults
+      @used_providers ||= (config_providers + default_providers).uniq
     end
 
     # @return [Array] - returns a list of providers that should always be loaded
--- a/spec/unit/pool_manager_spec.rb
+++ b/spec/unit/pool_manager_spec.rb
@@ -75,6 +75,84 @@ EOT
     end
   end
 
+  describe '#used_providers' do
+    context 'with no named providers' do
+      let(:config) { YAML.load(<<-EOT
+---
+:config:
+:pools:
+  - name: '#{pool}'
+    size: 1
+    provider: 'spoof'
+    EOT
+      )
+      }
+      it do
+        result = ["dummy"]
+        expect(subject.used_providers).to eq(result)
+      end
+    end
+    context 'with one named provider without a provider_class key' do
+      let(:config) { YAML.load(<<-EOT
+---
+:config:
+:providers:
+  :mock:
+:pools:
+  - name: '#{pool}'
+    size: 1
+    provider: 'spoof'
+    EOT
+      )
+      }
+      it do
+        result = ["mock", "dummy"]
+        expect(subject.used_providers).to eq(result)
+      end
+    end
+
+    context 'with one named provider with a provider_class key' do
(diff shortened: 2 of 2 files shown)


lib/vmpooler/pool_manager.rb
```diff
@@ -619,8 +619,24 @@ def load_used_providers
...
+            config_provider_name
+          end
+        end.compact.uniq
+      else
+        config_providers = []
+      end
+      # create an array of providers specified by pools
       pools = config[:pools] || []
-      @used_providers ||= (pools.map { |pool| pool[:provider] || pool['provider'] }.compact + default_providers).uniq
+      pool_provider_names = pools.map { |pool| pool[:provider] || pool['provider'] }.compact.uniq
```
sbeaulie, line 637: not sure this list is useful. It would get provider names from the pools but these normally point to providers listed in the providers hash.

genebean: You have a point. Let me see if any tests blow up without it

genebean: Seems okay, pushed

lib/vmpooler/pool_manager.rb
```diff
@@ -619,8 +619,21 @@ def load_used_providers
...
+        config_providers = config_provider_names.map do |config_provider_name|
+          if config[:providers][config_provider_name] && config[:providers][config_provider_name]['provider_class']
+            config[:providers][config_provider_name]['provider_class']
+          else
+            config_provider_name
+          end
+        end.compact.uniq
+      else
+        config_providers = []
+      end
```
sbeaulie, line 634: Nice use of map, which made me think we can one line it with the .map() for hashes which return an array
```suggestion
 config_providers = config[:providers].map { |k,v| v['provider_class'] || k.to_s }.compact.uniq
```

sbeaulie (changes requested):

lib/vmpooler/pool_manager.rb
genebean: causes `undefined method `[]' for nil:NilClass` to be thrown on basically every test

genebean: I think I see why...

genebean: nm. I think there are too many places that the key might not exist to work with that map you posted

sbeaulie: config_providers = config[:providers].map { |k,v| v['provider_class'] || k.to_s }.compact.uniq if config[:providers]

sbeaulie: just missing the case where its not set (which is probably a lot of tests)

sbeaulie: mabe keep the if config[:providers] else blocks

sbeaulie (approved):

spec/unit/pool_manager_spec.rb
```diff
@@ -75,6 +75,68 @@
...
+  :foo:
+:pools:
+  - name: '#{pool}'
+    size: 1
+    provider: 'spoof'
+    EOT
+      )
+      }
+      it do
+        result = ["mock_mock", :foo, "dummy"]
```
sbeaulie, line 134: we dont want symbols here as it wont load the files that way. 
if you keep your above block, you need to change config_provider_name.to_s

spec/unit/pool_manager_spec.rb
```diff
@@ -75,6 +75,68 @@
...
+        expect(subject.used_providers).to eq(result)
+      end
+    end
+  end
+
```
sbeaulie, line 139: add case where providers is not in the config and it should return just the default "dummy"

lib/vmpooler/pool_manager.rb
genebean: sadly, still blows up

spec/unit/pool_manager_spec.rb
genebean: will do

spec/unit/pool_manager_spec.rb
genebean: good call

lib/vmpooler/pool_manager.rb
genebean: I am still going to look at this a little more though, I like what you've started

spec/unit/pool_manager_spec.rb
genebean: done

spec/unit/pool_manager_spec.rb
genebean: done

lib/vmpooler/pool_manager.rb
genebean: It seems there are too many places things can be blank to shrink this down more. I think I am just going to leave it for now.

PR #472, opened 2021-12-23
(maint) Fix EXTRA_CONFIG merge behavior

sbeaulie:
Before this change if an extra config file had new keys they would get
merged to the main config but if it contained an existing key, like
'providers' it would overwrite the original config.
Adding a library 'deep_merge' to do a more natural merge, where existing keys
get sub-elements added together and arrays are combined like for the
pool configuration.
Adding spec tests around EXTRA_CONFIG as they were missing, by adding
and testing two new extra_config.yaml fixture files

### DIFF
--- a/lib/vmpooler/providers/base.rb
+++ b/lib/vmpooler/providers/base.rb
@@ -242,7 +242,7 @@ module Vmpooler
         # returns
         #   nil when successful. Raises error when encountered
         def create_template_delta_disks(_pool)
-          raise("#{self.class.name} does not implement create_template_delta_disks")
+          puts("#{self.class.name} does not implement create_template_delta_disks")
         end
 
         # inputs
--- a/lib/vmpooler.rb
+++ b/lib/vmpooler.rb
@@ -3,6 +3,7 @@
 module Vmpooler
   require 'concurrent'
   require 'date'
+  require 'deep_merge'
   require 'json'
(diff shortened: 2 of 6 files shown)


lib/vmpooler/providers/base.rb
```diff
@@ -242,7 +242,7 @@ def vm_exists?(pool_name, vm_name)
         # returns
         #   nil when successful. Raises error when encountered
         def create_template_delta_disks(_pool)
-          raise("#{self.class.name} does not implement create_template_delta_disks")
+          puts("#{self.class.name} does not implement create_template_delta_disks")
         end
```
genebean, line 246: should the delta disks stuff move to the vsphere provider?

genebean (approved):
LGTM. Feel free to merge if you are ready.

lib/vmpooler/providers/base.rb
sbeaulie: the actual implementation is in vsphere, but the manager does call it once per pool (and then marks it as done in redis). Not raising the error in the base class is what I have done for other vsphere specific method calls.

Another option would be to swallow the raised error or check that  the provider is vsphere before running it , but that would hide the base class ,ethod and I dont like it as much.

genebean: Makes sense

PR #475, opened 2021-12-30
(DIO-2833) Connect domain settings to pools, create v2 API

genebean:
### API V2

There are changes needed to support domains being configurable per pool that will require breaking changes to the API. For this reason a V2 API is being introduced as part of this PR. The V2 API is intended to be minimally different than the V1 with the following key differences:

- hostnames in the JSON response will be fully qualified if a provider-level or as as fallback a global domain is specified. In the V1 API they were always bare hostnames (sans domain).
- the top-level key domain that is included in V1 JSON responses will no longer be included since each host could be from a different domain.


spec/unit/pool_manager_spec.rb
```diff
@@ -2704,6 +2774,7 @@
       allow(subject).to receive(:check_snapshot_queue)
       allow(subject).to receive(:check_pool)
       allow(subject).to receive(:check_ondemand_requests)
+      allow(subject).to receive(:get_domain_for_pool).with(pool).and_return('example.com')
```
sbeaulie, line 2777: I think the spec error is due to this with(), it could be removed to accept any parameter.

lib/vmpooler/pool_manager.rb
```diff
@@ -658,6 +659,21 @@ def get_provider_for_pool(pool_name)
       $providers[provider_name]
     end
 
+    # @param pool_name [String] - the name of the pool
+    # @return [String] - domain name for pool, if set
+    def get_domain_for_pool(pool_name)
+      provider = get_provider_for_pool(pool_name)
```
sbeaulie, line 665: I think this can return `nil` too

spec/unit/pool_manager_spec.rb
```diff
@@ -1668,6 +1668,76 @@
...
+      let(:config) { YAML.load(<<~EOT
+        ---
+        :config:
+        :providers:
+          :mock_provider:
+            domain: m.example.com
+        :pools:
+          - name: '#{pool}'
+            size: 1
+            provider: #{provider_name}
```
sbeaulie, line 1727: another case could be added where `provider` is not set

genebean:
I will be resuming this work today

lib/vmpooler/pool_manager.rb
genebean: Thanks, I am adding a new line to bail early if the respone is nil

spec/unit/pool_manager_spec.rb
genebean: Nice catch! That did indeed fix the tests

spec/unit/pool_manager_spec.rb
genebean: I just added such a test a little above here and am about po push the code.

lib/vmpooler/pool_manager.rb
```diff
@@ -361,35 +362,39 @@ def find_unique_hostname(pool_name)
...
+      fqdn
     end
 
-    def check_dns_available(vm_name, domain = nil)
-      # Query the DNS for the name we want to create and if it already exists, mark it unavailable
-      # This protects against stale DNS records
-      vm_name = "#{vm_name}.#{domain}" if domain
+    # Query the DNS for the name we want to create and if it already exists, mark it unavailable
+    # This protects against stale DNS records
+    def check_dns_available(vm_name)
```
sbeaulie, line 392: we might not want to do this check anymore? In GCE if the DNS already exist upon vm creation, we replace it with the new IP. 

In this legacy code here, it would see that the dns already exists, and re-generate a new `spicy-proton` name. It is still applicable for vsphere though, so not sure what we might want to do.

genebean: We could add in configuration that allows a provider to opt out of this check... what do you think about that?

sbeaulie: yes that would be great. like if provider.skip_dns_avalaible_check == true then we don't run this method in GCE (additionally I think we pay for DNS queries).

As long as we can get the provider in this context, that would work well.

lib/vmpooler/api/v2.rb
```diff
@@ -0,0 +1,429 @@
...
+              template_backends.delete(first)
+              template_backends.unshift(first)
+            end
+          end
+
+          checkoutlock.synchronize do
+            template_backends.each do |template_backend|
+              vms = backend.smembers("vmpooler__ready__#{template_backend}")
+              next if vms.empty?
+
```
yachub, line 74: Looks like rubocop isn't happy about this whitespace:
```suggestion

```

yachub (approved):

PR #476, opened 2022-01-19
(DIO-2833) Update dev tooling and related docs

genebean:
These are changes that were found to be needed while working on the DIO-2833 ticket.

Once approved, we will need to adjust the suite of required tests in the branch protection rule for `main`.


yachub (approved):

yachub:
Followed the new docs and the new docker-compose environment worked great 😄

PR #489, opened 2022-03-30
(maint) Fix deprecation warning for redis ruby library

sbeaulie:
The syntax for pipelined has changed and the old syntax will be removed in v5.0.0
Fixing the syntax now since the block syntax has been supported for a while now.


sbeaulie:
spec tests failure need to look into it

yachub (approved):

PR #492, opened 2022-04-02
simplify gemfile lock generation and add snyk

LivingInSyn:
This PR does two things:

* simplifies and speeds up the generation/updating of the `Gemfile.lock`
* Adds snyk scanning when pushing to main

### DIFF
--- a/update-gemfile-lock.sh
+++ b/update-gemfile-lock.sh
@@ -3,4 +3,4 @@
 docker run -it --rm \
   -v $(pwd):/app \
   $(grep ^FROM docker/Dockerfile |cut -d ' ' -f2) \
-  /bin/bash -c 'apt-get update -qq && apt-get install -y --no-install-recommends make && cd /app && gem install bundler && bundle install --jobs 3 && bundle update; echo "LOCK_FILE_UPDATE_EXIT_CODE=$?"'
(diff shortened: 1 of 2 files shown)


update-gemfile-lock.sh
```diff
@@ -3,4 +3,4 @@
 docker run -it --rm \
   -v $(pwd):/app \
   $(grep ^FROM docker/Dockerfile |cut -d ' ' -f2) \
-  /bin/bash -c 'apt-get update -qq && apt-get install -y --no-install-recommends make && cd /app && gem install bundler && bundle install --jobs 3 && bundle update; echo "LOCK_FILE_UPDATE_EXIT_CODE=$?"'
+  /bin/bash -c 'cd /app && gem install bundler && bundle lock --update; echo "LOCK_FILE_UPDATE_EXIT_CODE=$?"'
```
genebean, line 6: The stock container that is used as part of this needs `make` to be installed before bundler is run.

genebean (changes requested):

update-gemfile-lock.sh
LivingInSyn: I tested it and it worked fine, can you share somewhere that it requires `make`?

yachub:
Closing as we're migrating away from Snyk anyways.

PR #551, opened 2023-02-07
Implement DNS Plugins

yachub:


lib/vmpooler/dns/base.rb
```diff
@@ -0,0 +1,85 @@
...
+          @config = config
+          @logger = logger
+          @metrics = metrics
+          @redis = redis_connection_pool
+          @dns_plugin_name = name
+
+          @dns_options = options
+
+          logger.log('s', "[!] Creating dns plugin '#{name}'")
+          # Your code goes here...
```
e-gris, line 25: Remove this.

lib/vmpooler/dns/base.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+
+        def name
+          @dns_plugin_name
+        end
+
+        def get_ip(vm_name)
+          @redis.with_metrics do |redis|
+            ip = redis.hget("vmpooler__vm__#{vm_name}", 'ip')
+            return ip
```
e-gris, line 61: Redundant assignment.

```ruby
@redis.with_metrics do |redis|
   redis.hget("vmpooler__vm__#{vm_name}", 'ip')
end
```

lib/vmpooler/dns/base.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        # returns
+        #   Array[String] : Array of pool names this provider services
+        def provided_pools
+          list = []
+          @config[:pools].each do |pool|
+            list << pool['name'] if pool['dns_config'] == name
+          end
+          list
+        end
```
e-gris, line 73: Use select/map. Also not sure what `name` is here, so I'll leave it be.

```ruby
def provided_pools
  @config[:pools].select { |pool| pool['dns_config'] == name }.map { |pool| pool['name'] }
end

File: lib/vmpooler/pool_manager.rb [hunk 4]
```diff
@@ -106,6 +113,10 @@ def _check_pending_vm(vm, pool, timeout, provider)
 
     def remove_nonexistent_vm(vm, pool, redis)
       redis.srem("vmpooler__pending__#{pool}", vm)
+      dns_plugin = get_dns_plugin_class_for_pool(pool)
+      domain = get_dns_plugin_domain_for_pool(pool)
+      fqdn = vm + '.' + domain
```
e-gris, line 118: Use interpolation.

e-gris (commented):
Only comments on Ruby style or idiomatic usage. Feel free to take or ignore.

Also, generally, avoid method variables like `name`, Even the method comments have suggestions for better choices.

I've shown a few examples but didn't flag them all.

lib/vmpooler/api/v2.rb
```diff
@@ -24,6 +65,56 @@ def get_template_aliases(template)
...
+            pickup = Pickup.new(weighted_pools)
+            count.to_i.times do
+              selection << pickup.pick
+            end
+          else
+            count.to_i.times do
+              selection << template_backends.sample
+            end
+          end
+        end
```
e-gris, line 113: Better form is to avoid multiple if/else layers using guardian clauses. This increases code maintainability by reducing code paths. Here's an example:

```ruby
def evaluate_template_aliases(template, count)
  [ ... ]
  aliases =  get_template_aliases(template)
  return [] unless aliases

  selection = []
  template_backends += aliases
  weighted_pools = get_pool_weights(template_backends)

  if weighted_pools.count > 1 && weighted_pools.count == template_backends.count
    pickup = Pickup.new(weighted_pools)
    count.to_i.times do
      selection << pickup.pick
    end
    return selection
  end

  count.to_i.times do
    selection << template_backends.sample
  end

  return selection
end

File: lib/vmpooler/api/v2.rb [hunk 6]
```diff
@@ -146,6 +271,254 @@ def atomically_allocate_vms(payload)
...
+            backend.hget("vmpooler__vm__#{vmname}", 'token:user')
+            backend.hget("vmpooler__vm__#{vmname}", 'template')
+            jenkins_build_url, user, poolname = backend.exec
+            poolname = poolname.gsub('.', '_')
+  
+            if user
+              user = user.gsub('.', '_')
+            else
+              user = 'unauthenticated'
+            end
```
e-gris, line 302: Idiomatically, this would be better expressed as:

```ruby
user = if user
  user.gsub('.', '_')
else
  'unathenticated'
end
```

Also reusing `user` should be avoided. I would likely do something like:

```ruby
jenkins_build_url, default_user, poolname = backend.exec
user = if default_user
             default_user.gsub('.', '_')
           else
             'unauthenticated'
           end
```

lib/vmpooler/api/v2.rb
```diff
@@ -146,6 +271,254 @@ def atomically_allocate_vms(payload)
...
+            metrics.increment("user.#{user}.#{operation}.#{poolname}")
+  
+            if jenkins_build_url
+              if jenkins_build_url.include? 'litmus'
```
e-gris, line 306: Nested 'if's can be refactored. I think just something like:

```ruby
return unless jenkins_build_url

if jenkins_build_url.include? 'litmus'
  metrics.increment("usage_litmus.#{user}.#{operation}.#{poolname}")
  return
end

url_parts = jenkins_build_url.split('/')[2..-1]
[ .... ]
```

e-gris (commented):
Some more purely stylistic/maintainability comments; certainly ignorable for now. I called out just a couple of examples.

Nested 'ifs' and variable mutations are great places for bugs to hide and are often easily refactored into something more stable

lib/vmpooler/dns/base.rb
yachub: Thanks, very concise! `name` is just name of the plugin that should be overridden in child classes, for example 'gcp': https://github.com/puppetlabs/vmpooler-dns-gcp/blob/63bf1a8060e6b46f8a387eecde1bec135b1ac2b2/lib/vmpooler/dns/gcp.rb#L43

lib/vmpooler/api/v2.rb
yachub: For all the comments in this file, everything here was just a straight copy/paste from the apiv1 to apiv2, since previously this v2 inherited from v1.

e-gris (approved):

e-gris (approved):

README.md
```diff
@@ -2,34 +2,66 @@
...
 
 VMPooler provides configurable 'pools' of instantly-available (pre-provisioned) and/or on-demand (provisioned on request) virtual machines.
 
 ## Usage
 
 At [Puppet, Inc.](http://puppet.com) we run acceptance tests on thousands of disposable VMs every day. VMPooler manages the life cycle of these VMs from request through deletion, with options available to pool ready instances, and provision on demand.
 
+### Migrating to v3
+
+Starting with the v3.x release, management of DNS records has implemented as DNS plugins, similar to compute providers. This means each pool configuration should be pointing to a configuration object in `:dns_config` to determine it's method of record management.
```
genebean, line 35:
```suggestion
Starting with the v3.x release, management of DNS records is implemented as DNS plugins, similar to compute providers. This means each pool configuration should be pointing to a configuration object in `:dns_config` to determine it's method of record management.
```

README.md
```diff
@@ -2,34 +2,66 @@
...
+
+For those using the global `DOMAIN` environment variable or global `:config.domain` key, this means records were not previously being managed by VMPooler (presumably managed via dynamic dns), so it's value should be moved to `:dns_configs:<INSERT_YOUR_OWN_SYMBOL>:domain` with the value for `dns_class` for the config set to `dynamic-dns`.
+
+For example, the following < v3.x configuration:
+
+```yaml
+:config:
+  domain: 'example.com'
+```
```
genebean, line 44: If this has not been done already, maybe a check could be added in vmpooler.rb that causes startup to fail if this setting is present.

docs/API-v2.md
```diff
@@ -12,6 +12,15 @@
 
 vmpooler provides a REST API for VM management.  The following examples use `curl` for communication.
 
+## Major change in V3 versus V2
```
genebean, line 15: This heading is wrong - you have not yet introduced a v3 api.... it kinda sounds like you should though. I will likely revisit this comment. My reading thus far is that you are introducing breaking changes which means you should be implementing them as a v3 api instead of modifying the v2 one. Removing the v2 api in addition to the v1 api is perfectly acceptable if that is the level of change desired, but you should not introduce breaking changes in a versioned api endpoint.

lib/vmpooler/api/reroute.rb
```diff
@@ -3,85 +3,69 @@
...
       end
 
       post '/vm/:hostname/snapshot/:snapshot/?' do
-        puts "DEPRECATION WARNING a client (#{request.user_agent}) called post /vm/:hostname/snapshot/:snapshot/? and got redirected to api_version=1, this behavior will change in the next major version, please modify the client to use v2 in advance"
         call env.merge('PATH_INFO' => "/api/v#{api_version}/vm/#{params[:hostname]}/snapshot/#{params[:snapshot]}")
       end
 
       put '/vm/:hostname/disk/:size/?' do
-        puts "DEPRECATION WARNING a client (#{request.user_agent}) called put /vm/:hostname/disk/:size/? and got redirected to api_version=1, this behavior will change in the next major version, please modify the client to use v2 in advance"
         call env.merge('PATH_INFO' => "/api/v#{api_version}/vm/#{params[:hostname]}/disk/#{params[:size]}")
```
genebean, line 69: Can we just kill this off and simply use the versioned endpoints as part of the breaking changes already being introduced?

genebean (changes requested):
So, I am only part of the way through, but it seems that this PR introduces breaking changes within the versioned API endpoint `api/v2`... that is not okay. Breaking changes need to be implemented as a new api version. I think what you are doing here api wise need to be done as part of `api/v3`. It is fine if you kill of the v1 and v2 endpoints as part of this work too.

docs/API-v2.md
yachub: Oops I got some wires crossed here when updating the readme (Referring to V2 and V3 of vmpooler gem itself) vs the api doc (Referring to the V2 and V3 versioned endpoints).

README.md
yachub: Sure that should be an easy add.

yachub: Done

spec/unit/dns/base_spec.rb
genebean, line 1: Please add a new line at the end of the file

genebean (changes requested):

lib/vmpooler/dns.rb
```diff
@@ -0,0 +1,99 @@
...
+    # Returns the domain for the specified pool
+    #
+    # @param config [String] - the full config structure
+    # @param pool_name [String] - the name of the pool
+    # @return [String] - domain name for pool, which is set via reference to the dns_configs block
+    def self.get_domain_for_pool(config, pool_name)
+      pool = config[:pools].find { |p| p['name'] == pool_name }
+      pool_dns_config = pool['dns_plugin']
+      dns_configs = config[:dns_configs].keys
+      pool_domain = ''
```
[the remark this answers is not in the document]
yachub: Ya these 2 must have been an earlier iteration that I forgot to fix, will do

spec/integration/api/v2/vm_spec.rb
```diff
@@ -341,6 +355,30 @@ def app()
...
+      #     ok: true,
+      #     pool1: {
+      #       hostname: "2#{vmname}.one.example.com"
+      #     }
+      #   }
+
+      #   expect(last_response.body).to eq(JSON.pretty_generate(expected))
+
+      #   expect(pool_has_ready_vm?('pool1', vmname, redis)).to be false
+      # end
```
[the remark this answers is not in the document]
yachub: After re-reading this it actually is valid, I added it back in.

lib/vmpooler/pool_manager.rb
```diff
@@ -437,16 +445,25 @@ def _clone_vm(pool_name, provider, request_id = nil, pool_alias = nil)
...
           provider.create_vm(pool_name, new_vmname)
           finish = format('%<time>.2f', time: Time.now - start)
+          $logger.log('s', "[+] [#{pool_name}] '#{new_vmname}' cloned in #{finish} seconds")
+          $metrics.timing("clone.#{pool_name}", finish)
+
+          $logger.log('d', "[ ] [#{pool_name}] Obtaining IP for '#{new_vmname}'")
+          ip_start = Time.now
+          ip = provider.get_vm_ip_address(new_vmname, pool_name)
+          ip_finish = format('%<time>.2f', time: Time.now - ip_start)
+          $logger.log('s', "[+] [#{pool_name}] Obtained IP for '#{new_vmname}' in #{ip_finish} seconds")
```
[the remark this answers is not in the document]
yachub: Ya since pool manager doesn't currently implement any tracing, probably better as a separate task. I opened https://puppet.atlassian.net/browse/POOLER-215 for that.

genebean (approved):
Please don't merge until after squashing (I know you are working on it). Adding a label to ensure this.

yachub:
Just working on squashing commits now

PR #555, opened 2023-03-02
(maint) Use timeout builtin to TCPSocket when opening sockets.

isaac-hammes:


yachub:
To remove Timeout usage per https://github.com/mperham/connection_pool

> WARNING: Don't ever use Timeout.timeout in your Ruby code or you will see occasional silent corruption and mysterious errors. The Timeout API is unsafe and cannot be used correctly, ever. Use proper socket timeout options as exposed by Net::HTTP, Redis, Dalli, etc.

yachub (approved):

PR #561, opened 2023-03-15
(RE-15162) Update Redis gem to version 5.0.

### LINKED ISSUE #603: Redis 5.x Deprecations
This issue has been copied from Jira at https://puppet.atlassian.net/browse/POOLER-216

isaac-hammes:
This should fix deprecations that were removed in https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#500

Fixes #603


lib/vmpooler.rb
```diff
@@ -206,8 +206,7 @@ def self.redis_connection_pool(host, port, password, size, timeout, metrics, red
   end
 
   def self.new_redis(host = 'localhost', port = nil, password = nil, redis_reconnect_attempts = 10)
-    Redis.new(host: host, port: port, password: password, reconnect_attempts: redis_reconnect_attempts, reconnect_delay: 1.5,
-              reconnect_delay_max: 10.0)
+    Redis.new(host: host, port: port, password: password, reconnect_attempts: redis_reconnect_attempts)
```
yachub, line 209: Can we pass in the `reconnect_delay` and `reconnect_delay_max` options into `reconnect_attempts` instead of dropping them?

Not sure of the format off the top of my head, but it looks like that's the new format based on https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#500
> Removed reconnect_delay_max and reconnect_delay, you can pass precise sleep durations to reconnect_attempts instead.

isaac-hammes: Sorry, I just saw this. Yes, we can do that. That is what I was planning to do. It looks like I need to update it in vmpooler-deployment. What do you think the reconnect attempts should look like? Maybe a progression of waiting longer and longer each time? [0, 0.1, 0.2, 0.4, 0.8, 1.6, 3.2, 6.4, 12.8, 25.6, 51.2, 102.4, 204.8]? It's set for 40 reconnect attempts now.

yachub (approved):

Issue #609, opened 2023-08-14
Log reason connection on port 22 of a failed VM


yachub:
Currently, when vmpooler waits for a VM to become responsive, it tries to connect to a socket on port 22. If it fails, no error information is saved or logged anywhere.
Change this to log this error somewhere.

isaac-hammes:
2023-08-18 06:04:07.956 | 2023-08-18T13:04:07.724584214Z stdout F [2023-08-18 13:04:07] [!] [fedora-36-x86_64] 'snowy-unison' marked as 'failed' after 15 minutes with ssh error: Connection timed out - Broken pipe
-- | --

PR #611, opened 2023-08-16
(POD-10) Log reason for failed VM checks.

isaac-hammes:
Fixes https://github.com/puppetlabs/vmpooler/issues/609


isaac-hammes:
I deleted the vm_ready? method from vmpooler because I do not believe it is actually being used anywhere, only the provider's method is being used.

yachub (approved):

PR #612, opened 2023-08-16
Update changelog

yachub:
I have a suspicion that the `Merge branch 'main' into update_redis` at https://github.com/puppetlabs/vmpooler/commit/113ca2dacbb8fed80a28118c28f353b80108d9c6 might be causing some issues with the changelog generator. However, the issue for that PR is still in the changelog so it's easy to see that the change happened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,8 @@
 # Changelog
 
-## [3.3.0](https://github.com/puppetlabs/vmpooler/tree/3.3.0) (2023-08-15)
+## [3.3.0](https://github.com/puppetlabs/vmpooler/tree/3.3.0) (2023-08-16)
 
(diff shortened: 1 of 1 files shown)


e-gris (commented):
Looks like Isaac's note was deleted in this change. Was that intentional?

yachub:
> Looks like Isaac's note was deleted in this change. Was that intentional?

Yes, I threw up the PR then had to run to a meeting. I have a suspicion that the `[Merge branch 'main' into update_redis]` at https://github.com/puppetlabs/vmpooler/commit/113ca2dacbb8fed80a28118c28f353b80108d9c6 might be causing some issues with the changelog generator. However, the issue for that PR is still in the changelog so it's easy to see that the change happened.

isaac-hammes (approved):

PR #621, opened 2023-08-23
(maint) Fix bugs from redis and timeout notification updates.

isaac-hammes:


lib/vmpooler/pool_manager.rb
```diff
@@ -129,6 +129,8 @@ def fail_pending_vm(vm, pool, timeout, timeout_notification, redis, exists: true
       already_timed_out = time_since_clone > timeout
       timing_out_soon = time_since_clone > timeout_notification && !redis.hget("vmpooler__vm__#{vm}", 'timeout_notification')
 
+      return true if !already_timed_out && !timing_out_soon
```
yachub, line 132: If it would be easy to account for this in a test case, then might be worth it? Otherwise, if a timely fix is needed, then that's fine.

isaac-hammes: Yeah, I don't think it should be hard. I'll get a test case added for it. Thanks for the feedback.

isaac-hammes: It doesn't seem to be super urgent as I am unaware of anything actually failing because of it. I think it is just spamming the logs.

PR #680, opened 2025-02-19
Upgrade file check logic for newer ruby versions

trvs-sdlr:


isaac-hammes:
Looks good, but could you bump jruby to jruby-9.4.10.0 in the [test](https://github.com/puppetlabs/vmpooler/blob/main/.github/workflows/testing.yml) action?

